PR #8, opened 2016-06-30
Update HomeView.js

mattboutet:
the require of the Duck.jpg asset fails on linux because case sensitive.


devinivy:
Fixed in a170425621b8c2a4ec04f5e4a3ab500769eddca8 (by changing the case of the file itself rather than the `require()`).

Issue #30, opened 2016-08-08
IE 11 Fails To Load Applications


geoff5802:
A issue inherited from "React-redux-starter-kit", details can be seen here: https://github.com/davezuko/react-redux-starter-kit/issues/785

The PR for that issue can be seen here - https://github.com/creatorkuang/react-redux-starter-kit/commit/dcb1c88c03d65b79c8ec70d63d2ef3e2236ffa84

We just need to update config/main.js to reflect their changes and it seems to solve the issue at hand. # @

devinivy:
Resolved in #31

PR #31, opened 2016-08-09
fixing IE 11 bug

geoff5802:
bug fix is based on recommendation in related issue, tested in separate fork.


devinivy:
Resolves #30

PR #127, opened 2017-02-16
reverting several package updates due to breakage

geoff5802:


package.json
```diff
@@ -89,28 +89,28 @@
     "babel-preset-stage-0": "^6.3.13",
     "babel-register": "^6.3.13",
     "babel-runtime": "^6.3.19",
-    "better-npm-run": "0.0.11",
+    "better-npm-run": "0.0.14",
     "css-loader": "^0.25.0",
     "cssnano": "^3.3.2",
-    "debug": "^2.2.0",
+    "debug": "^2.4H.0",
```
devinivy: What is this?

package.json
```diff
@@ -89,28 +89,28 @@
     "babel-preset-stage-0": "^6.3.13",
     "babel-register": "^6.3.13",
     "babel-runtime": "^6.3.19",
-    "better-npm-run": "0.0.11",
+    "better-npm-run": "0.0.14",
```
devinivy: No problem that some of these versions went up, and some went down?

geoff5802: some went down because I'm reverting previously merged package updates.

geoff5802:
Breaking updates includes those to the following packages:

- History
- Hapi
- React-router
- webpack
- sass-loader
- postcss-loader

bmleight:
@geoff5802 looks ok to me.  Packages reverting to older versions also caught my eye.  My understanding is that those versions had already been merged into master, but you found they break some shit.  Is that correct?  I just want to get to the bottom of how they got merged in so we can work out a flow to prevent this going forward

geoff5802:
@bmleight my initial process was to merge the the greenkeeper PR's, the pull them down to see if they broke stuff, and I was doing this in batches (as there were a ton of PR's to get through). I then started making all the changes locally and just deleted the greenkeeper pr's because the cycle time was too long to continue with my previous process.

PR #171, opened 2017-04-10
Upgrade to webpack v2

devinivy:
Along with upgrading to webpack v2, many other deps are also upgraded.  History and react-router are still tied to each other– actually waiting on react-router-redux v5 before upgrading all three of those together.


geoff5802:
i find this very exciting.

PR #200, opened 2017-07-15
Just ignorin the package-lock

wswoodruff:


devinivy:
I think it's really up to the user if they want to use package-lock or not– to ignore it is to say that strangeluv officially discourages use of npm5+.  I would suggest just doing the ignore in your project instead!  @bmleight @geoff5802 thoughts?  @wswoodruff what do you think?

wswoodruff:
My gripe was when pulling in new flavors I was getting errors after remembering to run `npm install` after pulling them in

devinivy:
Try it on npm v5.1! https://github.com/npm/npm/releases/tag/v5.1.0

wswoodruff:
Ok, I did `npm install -g npm@latest` and now I'm at npm -v 5.3.0. loL wish me luck!

PR #210, opened 2018-03-22
Material UI 1.x.x Recipe

geoff5802:
- Add two new packages, **material-ui** and **styled-components**. The other package updates are incidental and should be backed out of this pr.
- Create mui theme file, and implement on core layout
- Create color file so application colors can be centralized 
- Add buttons to header in three different flavors, straight out the box, wrapped in HOC withStyles, and using styled-components package.


rbkh:
*recipe ;)

geoff5802:
I was using this work as a reference today and realized this PR no longer conforms to the latest Material UI release.

**Ex.**
`const purple = require('material-ui/colors/purple').default;`
**is now**
`const purple = require('@material-ui/core/colors/purple').default;`

PR #215, opened 2018-04-16
Fix tests

bmleight:
Did the following to get existing tests passing again:

* Upgraded various dependencies
* Switched from phantom to puppeteer
* Fixed tests failing due to routing changes


package.json
```diff
@@ -78,6 +78,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "ajv": "^6.4.0",
```
devinivy: Where does ajv come into play?

package.json
```diff
@@ -110,7 +111,8 @@
     "react-dom": "^16.0.0",
     "react-redux": "^5.0.2",
     "react-router-dom": "^4.2.2",
-    "react-router-redux": "5.0.0-alpha.6",
+    "react-router-redux": "5.0.0-alpha.9",
+    "react-test-renderer": "^16.3.1",
```
devinivy: Should this be a dev dep?

package.json
bmleight: you're quick!  I was just reviewing and removing this.  Not _exactly_ sure how I ended up getting it in here.  From early efforts.

PR #217, opened 2018-04-24
Recipe initial load action creator

bmleight:
This implementation is based on [Redux's docs on reducing boilerplate](https://redux.js.org/recipes/reducing-boilerplate#async-action-creators).  The approach is to implement an API [middleware](https://redux.js.org/glossary#middleware) that listens for a new `callAPI` action type, shoots off the API request as well as the corresponding `start`, `success`, and `fail` actions.

This is not meant to be a perfect implementation of this, but as an example to illustrate this approach.


devinivy:
What are your impressions?

PR #218, opened 2018-04-26
Add emotion example, couple changes to styled-components example

wswoodruff:
The border changes color every time the component is rendered, but this is to show that your state can affect your styles very easily (which is good!)


package.json
```diff
@@ -109,6 +110,7 @@
     "postcss-loader": "^2.1.2",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
+    "react-emotion": "^9.1.2",
```
wswoodruff: Both the `emotion` and `react-emotion` libraries are required for `styled-componets`-like syntax. Checkout how the 2 files differ!! 🎉

src/components/material/button-emotion.js
```diff
@@ -0,0 +1,54 @@
+const React = require('react');
```
wswoodruff: Checkout how much this file differs from the styled-components example!! 🎉

PR #219, opened 2018-05-11
Recipe login

rbkh:
This recipe has all the essential functionality for basic user authentication:

- Sign up with name, email, password to create an account
- Login and logout
- "Remember Me" functionality 
- Forget password / reset password flow (requires email service integration to userpal to function)
- Basic form validation (is email a real email address, do the passwords match), plus validation error and server error handling
- Auth initializer to manage `rememberMe` and `strange-auth` auth status on init
- Example of an authenticated route (`dashboard`) that you can only access when logged in


rbkh:
note - travis is failing cause it's looking for `counter` example (which I removed)

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
+const StrangeAuth = require('strange-auth');
+const WebClient = require('../utils/web-client');
+const History = require('../wiring/history');
+const AuthAct = require('../action-types/auth');
```
wswoodruff: For utils, wiring, and action-types you shouldn't need to do relative paths here, I think you should be able to just do

const WebClient = require('utils/web-client');
const History = require('wiring/history');
const AuthAct = require('action-types/auth');

devinivy: I personally like seeing the relative paths (esp when we're only one level deep, and not way down in `routes/`), but that jus me.

rbkh: I agree with Devin 😛but can do whatever the group wants!

wswoodruff: Don't worry about making a change for that =)

src/components/Header/index.js
```diff
@@ -1,18 +1,42 @@
...
+                <NavLink to='/dashboard' activeClassName={Classes.activeRoute}>
+                    Dashboard
+                </NavLink>
+                {' · '}
+                {!isAuthenticated &&
+                    <NavLink to='/sign-up' activeClassName={Classes.activeRoute}>
+                        Sign Up
+                    </NavLink>
+                }
+                {!isAuthenticated &&
```
bmleight: You might be able to use a [fragment](https://reactjs.org/docs/fragments.html) to avoid `!isAuthenticated` twice in a row.

src/components/Header/index.js
```diff
@@ -1,18 +1,42 @@
...
+                    <span> · </span>
+                }
+                {isAuthenticated ? <button onClick={logout}>Logout</button> : <NavLink to='/login' activeClassName={Classes.activeRoute}>Login</NavLink>}
```
bmleight: Going off the above, you could move the login NavLink up there, and declutter this line.

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
...
+                if (status !== 200) {
+                    const err = new Error('Bad login');
+                    err.info = data;
+                    return Promise.reject(err);
+                }
+
+                finalToken = data;
+
+                return WebClient.get('/user', {
+                    headers: { authorization: `Bearer ${finalToken}` }
```
bmleight: Maybe consolidate this with the same chunk of code at 208-210?  Could shove it in an internal function.

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
...
+                headers: { authorization: `Bearer ${finalToken}` }
+            });
+        }
+
+        else {
+
+            authPromise = WebClient.post('/login', { email, password }, { responseType: 'text' })
+            .then(({ data, status }) => {
+
+                if (status !== 200) {
```
bmleight: Is this needed?  I know it's in other projects, but I think non 200's get funneled to catch by axios.

src/actions/counter.js
```diff
@@ -1,36 +0,0 @@
-const CounterTypes = require('action-types/counter');
```
bmleight: Let's delete the counter's tests as well.  That'll probably get the PR to pass travis, which is a nice warm feeling.

src/initializers/auth.js
```diff
@@ -0,0 +1,54 @@
+const LocalStorageAvailable = require('utils/check-local-storage');
+const AuthActions = require('actions/auth');
+const AuthSelectors = require('selectors/auth');
+
+module.exports = (store) => {
+
+    if (!LocalStorageAvailable()) {
+
+        // This forces strange-auth auth status to be set
```
bmleight: What "auth status" gets set by this?  Reviewing [strange auth docs](https://github.com/BigRoomStudios/strange-auth#usage), are you referring to the actual `status` piece of state, or does this get `token: false` into `credentials`?  I have assumptions, but haven't totally tracked down what the purpose of this is for.  It feels like this is something the default state could cover?  Maybe passed in as [initial state in the auth reducer](https://github.com/BigRoomStudios/strange-auth/blob/master/src/reducer.js#L6)?

src/routes/dashboard/containers/Dashboard.js
```diff
@@ -0,0 +1,13 @@
+const Connect = require('react-redux').connect;
+const Dashboard = require('../components/Dashboard');
+
+const internals = {};
+
+internals.connect = Connect(
+    (state) => ({
+        firstName: state.auth.credentials.user.firstName
+    })
+
```
bmleight: extra?

src/routes/dashboard/index.js
```diff
@@ -0,0 +1,10 @@
+const Dashboard = require('./containers/Dashboard');
+const Authenticate = require('../auth').authenticate;
+
+module.exports = {
+
+    path: 'dashboard',
+    component: Authenticate(Dashboard),
+    exact: true
+
```
bmleight: extras?

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+            getFormValue: this.getFormValue.bind(this)
+        });
+    }
+
+    getFormValue(e) {
+
+        return e.target.value || '';
+    }
+
+    formField(field, value) {
```
bmleight: `setField`?  `setFormField`?

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+
+        const isBlurred = { ...this.state.isBlurred };
+        const field = ev.target.id;
+
+        isBlurred[field] = true;
+
+        this.setState({ isBlurred });
+    }
+
+    invalidEmail = () => {
```
bmleight: For functions that return a boolean, I like tacking on `is` to the beginning to make the function sound like a question.  Helps with readability.

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+
+        if ( this.state.isBlurred.email) {
+
+            if ( IsEmail(this.state.email) ) {
```
bmleight: extra spaces?  and above?

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+
+                return false;
+            }
+
+            return true;
+        }
+    }
```
bmleight: The fact that this can return nothing (and it's relied upon to not show the error message) isn't grabbing me well.  Maybe if this function is changed to something like `shouldShowErrorMessage`, and always returns a boolean value it would feel better?  I think it would also read better when used below.

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+
+    submit = (ev) => {
+
+        const { email } = this.state;
+
+        this.props.requestReset({ email });
```
bmleight: Should submitting be prevented if email is empty or invalid?

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,106 @@
...
+        super(props);
+
+        this.state = {
+            email: '',
+            isBlurred: {
+                email: false
+            }
+        };
+
+        this.strangeForm({
```
bmleight: This was my direction, so it's my fault, but strangeforms could be overkill for a simple form like this.  Curious what others think.

src/routes/index.js
```diff
@@ -1,14 +1,22 @@
...
+const ResetPasswordRoute = require('./reset-password');
 
 // Create routes
 module.exports = (store) => ([{
     path: '/',
     component: CoreLayout,
     childRoutes: [
         Home,
-        CounterRoute(store)
+        LoginRoute,
```
bmleight: Could be nice to have these all nested together

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
...
+                dispatch(actions.resetPasswordSuccess());
+                return History.push('/login');
+
+            })
+            .catch((err) => {
+
+                let errMessage = 'Unable to reset password. Please try again.';
+
+                if (typeof err.response !== 'undefined') {
+                    errMessage = err.message;
```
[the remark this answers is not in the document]
rbkh: thanks! fixed.

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
...
+        const strangeLogin = internals.strangeActions.login({ email, password, token });
+
+        return dispatch(strangeLogin)
+
+        .then(() => {
+
+            History.push('/dashboard');
+
+        })
+        .catch(::console.warn);
```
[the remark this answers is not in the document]
rbkh: axios complains about an uncaught error without this

src/actions/auth.js
rbkh: @devinivy This is Brian's question I'm not sure about! I'll dig a little into it, but figured I'd ping you too :)

rbkh: sounds like it's not needed 👌

src/actions/auth.js
```diff
@@ -0,0 +1,248 @@
...
+
+                return WebClient.get('/user', {
+                    headers: { authorization: `Bearer ${finalToken}` }
+                });
+            });
+        }
+
+        return authPromise.then(({ data, status }) => {
+
+            if (status !== 200) {
```
wswoodruff: Marking in case you decide to remove the other status check

src/initializers/auth.js
rbkh: Yep! I mean the actual `status` piece of state. So you're suggesting that we set initial status state in the auth reducer? But how would we know what that state should look like? 

The way Devin explained it to me is that there are kind of three things we might care about: if a user is authenticated, if a user is not authenticated, or whether we even care about authentication at all. The initial state might look different depending on project/user needs, so wouldn't we want to manage that logic in the initializer instead of defaulting to an initial state in the reducer? I could be 100% wrong about all this tho.

src/routes/forgot-password/components/ForgotPassword.js
rbkh: I think before I had it returning null (cause I agree, not returning something felt weird but didn't know if that was enough justification).

src/routes/forgot-password/components/ForgotPassword.js
rbkh: implemented some logic to disable button 👍

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,110 @@
...
+
+        return (this.state.isBlurred.email) && !IsEmail(this.state.email);
+    }
+
+    disableButton = () => {
+
+        const { email } = this.state;
+        const fieldHasValue = email !== '';
+
+        if (fieldHasValue && IsEmail(email)) {
```
bmleight: I think this could be simplified to 

```
return !IsEmail(email);
```

assuming `IsEmail` returns `false` for `''`.

This could just be me and my weird brain, but I felt slow following the logic here.  I think it's because I'm used to thinking about it as what validates the form, instead of when to disable the button (and similarly, in the disableButton function, I'd have the if statement define what makes it true instead of false).  I find it more straightforward to think about what criteria deem a valid form, and to negate that when determining if the button should be active.  Not tryin to say it's bad, just that it's different from my thought process.

src/routes/login/components/Login.js
```diff
@@ -0,0 +1,138 @@
...
+
+    setFormField(field, value) {
+
+        this.setState({ [field]: value });
+    }
+
+    fieldBlurred = (ev) => {
+
+        const isBlurred = { ...this.state.isBlurred };
+        const field = ev.target.id;
```
bmleight: Seems like this could be a potentially fragile area.  Someone changing the id of the element (which is kinda broad) might not realize it'll break the blur functionality.

src/routes/forgot-password/components/ForgotPassword.js
bmleight: similar feedback in forms below

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,110 @@
...
+        this.props.requestReset({ email });
+
+        ev.preventDefault();
+    }
+
+    render() {
+
+        return (
+
+            <div>
```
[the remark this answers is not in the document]
bmleight: also noticed in forms below

src/initializers/auth-token.js
```diff
@@ -0,0 +1,26 @@
+const WebClient = require('../utils/web-client');
```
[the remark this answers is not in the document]
bmleight: Talking about the whole file.  This looks like a reference of my body models(?) and `initializers/auth.js` is the new way that's actually used.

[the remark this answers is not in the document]
rbkh: oh I see what you mean

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+    return (dispatch) => {
+
+        dispatch(actions.resetPasswordRequest());
+
+        return WebClient.post('/users/reset-password', { email, newPassword, resetToken }, { responseType: 'text' })
+
+        .then(({ data, status }) => {
+
+            dispatch(actions.resetPasswordSuccess());
+            return History.push('/login');
```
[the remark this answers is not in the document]
rbkh: I agree that would be nice, consider it done.

src/components/Header/index.js
```diff
@@ -1,18 +1,54 @@
...
+                <button onClick={this.props.logout}>Logout</button>
+            </React.Fragment>
+        );
+    }
+
+    render() {
+
+        const { isAuthenticated } = this.props;
+
+        const renderNav = isAuthenticated ? this.renderAuthenticated() : this.renderNotAuthenticated();
```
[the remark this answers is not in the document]
rbkh: good feedback, thank you 👌

src/routes/login/components/Login.js
rbkh: I can add some comments explaining that the id is tied to the functionality if you think that'd help.

src/routes/forgot-password/components/ForgotPassword.js
rbkh: thanks, I simplified this one! 

Yeah, I'm definitely thinking about this as _when should the button be disabled_ more than _is this form valid_ (and to your other point, I guess it make sense to me to have the default return true because the the exception would be a disabled button, meaning the button will spend more time being disabled (true) than enabled (false)).

I'll get more opinions on this approach!

src/routes/forgot-password/components/ForgotPassword.js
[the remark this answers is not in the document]
rbkh: removed!

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+exports.registerUser = ({ email, password, firstName, lastName }) => {
+
+    return (dispatch) => {
+
+        dispatch(actions.registrationRequest());
+
+        const newUser = WebClient.post('/users', { email, password, firstName, lastName });
+
+        newUser
+
```
devinivy: Any reason for this big ole newline?

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+        type: AuthAct.REGISTRATION_FAILURE,
+        payload: errMessage
+    };
+};
+
+exports.registerUser = ({ email, password, firstName, lastName }) => {
+
+    return (dispatch) => {
+
+        dispatch(actions.registrationRequest());
```
devinivy: Doesn't matter so much for this PR, but I like passing the action payload through here for any reducers that may be interested.

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+        const strangeLogin = internals.strangeActions.login({ email, password, token });
+
+        return dispatch(strangeLogin)
+
+        .then(() => {
+
+            return History.push('/dashboard');
+
+        })
+        .catch(::console.warn);
```
devinivy: Woah!  This is old.  Is it from an example somewhere?  Would not suggest using the `::` syntax—it was popular for a while, but I don't think it has stuck.

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+    return (dispatch) => {
+
+        const strangeLogin = internals.strangeActions.login({ email, password, token });
+
+        return dispatch(strangeLogin)
+
+        .then(() => {
+
+            return History.push('/dashboard');
+
```
devinivy: Any reason for this big ole newline?

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+    return (dispatch) => {
+
+        dispatch(actions.requestResetRequest());
+
+        return WebClient.post('/users/request-reset', { email }, { responseType: 'text' })
+
+        .then(({ data, status }) => {
+
+            dispatch(actions.requestResetSuccess());
+            return History.push('/login');
```
devinivy: I think we'd dispatch react-router-redux's `push()` action.  I don't know what happens when we use that versus using history.

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+
+// StrangeAuth
+
+internals.strangeActions = StrangeAuth.makeActions({
+    login: ({ email, password, token }) => {
+
+        let authPromise;
+        let finalToken;
+
+        if (!!token) {
```
devinivy: `if (token)` would be OK too!

wswoodruff: Lmao that was my fault -- I was on a `!!` kick for a lil bit and we wrote this part together

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+                credentials: {
+                    token: finalToken,
+                    user: data
+                }
+            };
+        });
+    }
+});
+
+internals.getUser = (token) => {
```
devinivy: The is cool—also would have been alright to define this function inside `login()`.

src/actions/auth.js
devinivy: I remember that kick!  Casting to boolean is useful esp. when you want to return a boolean from a function for the consumer's benefit, or use a function that expects a boolean value.  We also use it inside react `render()` funcs because some falsy values actually could render!  For example, `0` is falsey but renders "0" into the dom, which could break something like `{userCount && <span>{userCount} users</span>}.

wswoodruff: 👌 Noice.

Using booleans in conditionals is veeeery delicate at times

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+            .then(({ data, status }) => {
+
+                finalToken = data;
+
+                return internals.getUser(finalToken);
+
+            });
+        }
+
+        return authPromise.then(({ data, status }) => {
```
devinivy: This branching/async logic could also be written,
```js
let getToken = Promise.resolve(token);

if (!token) {
    getToken = WebClient.post('/login', { email, password }, { responseType: 'text' })
        .then(({ data }) => data)
}

return maybeLogin.then((finalToken) => internals.getUser(finalToken)) ...
```

src/components/Header/index.js
```diff
@@ -1,18 +1,54 @@
...
-
-);
+module.exports = class Header extends React.Component {
+
+    static propTypes = {
+        logout: T.func.isRequired,
+        isAuthenticated: T.bool.isRequired
+    };
+
+    renderNotAuthenticated = () => {
```
devinivy: Are we using this syntax now? ¯\_(ツ)_/¯

src/components/Header/index.js
```diff
@@ -1,18 +1,54 @@
...
+
+        return (
+
```
devinivy: Why these spaces?

src/components/Header/index.js
```diff
@@ -1,18 +1,54 @@
...
+            <React.Fragment>
+                {' · '}
+                <NavLink to='/sign-up' activeClassName={Classes.activeRoute}>Sign Up</NavLink>
+                {' · '}
+                <NavLink to='/login' activeClassName={Classes.activeRoute}>Login</NavLink>
+            </React.Fragment>
+        );
+    }
```
devinivy: You can also just,
```js
    return <React.Fragment>
        {stuff}
    </React.Fragment>;
```

But if you like the extra parens, there's nothing wrong with that either!  Just expect to see both styles kickin around.

src/reducers/auth.js
```diff
@@ -0,0 +1,56 @@
...
+const Deeply = require('../utils/deeply');
+const RouterTypes = require('react-router-redux');
+
+const authReducer = StrangeAuth.makeReducer();
+
+module.exports = (state, action) => {
+
+    state = authReducer(state, action);
+
+    const { type } = action;
```
devinivy: These two lines can be combined into `const { type, payload } = actions`.

src/routes/dashboard/containers/Dashboard.js
```diff
@@ -0,0 +1,12 @@
+const Connect = require('react-redux').connect;
+const Dashboard = require('../components/Dashboard');
+
+const internals = {};
+
+internals.connect = Connect(
+    (state) => ({
+        firstName: state.auth.credentials.user.firstName
```
devinivy: A selector could be used here.

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,103 @@
...
+            email: '',
+            isBlurred: {
+                email: false
+            }
+        };
+
+        this.strangeForm({
+            fields: ['email'],
+            get: (someProps, field) => this.state[field],
+            act: this.setFormField.bind(this),
```
devinivy: Given that `get` is defined inline, `act` could as well!

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -0,0 +1,103 @@
...
+            getFormValue: this.getFormValue.bind(this)
```
devinivy: This could be defined inline too!  Actually you probably don't need to configure it at all—I think this implementation is strange-forms's default.

src/components/Header/index.js
wswoodruff: For anybody wondering what this syntax does, it sets a func on the class that automatically binds `this` to be the class

- Replacing the need for `this.myFunc = this._myFunc.bind(this)` in the constructor

src/actions/auth.js
```diff
@@ -0,0 +1,246 @@
...
+const History = require('../wiring/history');
+const AuthAct = require('../action-types/auth');
+
+const internals = {};
+
+const actions = exports;
+
+// New User Registration
+
+exports.registrationRequest = () => {
```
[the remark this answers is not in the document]
rbkh: thanks i'll make this update!

src/components/Header/index.js
rbkh: I personally prefer arrow functions because it's tidier and less repetitive; what's the advantage of binding in the constructor? Performance-wise it's about the same, correct?

devinivy: I agree that it's more concise!  And I'm not worried about performance.  My issues with it are that

 - it's not yet standard javascript, and I'm not sure if it will be.
 - it's really weird because in all other situations an "arrow function" specifically does _not_ have its own `this`!  This is really confusing.
 - it's more explicit to bind for the purposes passing a method to props.  I think we sometimes forget _why_ the binding is necessary in the first place.
 - there are other options even _aside_ from binding—it's no longer against our lint to use arrow functions in `render()`, for example.  So we could do `prop={() => this.myMethod()}` in many cases, and author our methods in the typical (non-arrow) way.

src/routes/forgot-password/components/ForgotPassword.js
rbkh: @devinivy would love to get your opinion (this convo got buried) !

devinivy: Is this primarily a question about how it should behave, or how it should be implemented?

src/components/Header/index.js
rbkh: I kinda like the extra parens (with the extra newlines I like to toss in 😛)

src/routes/forgot-password/components/ForgotPassword.js
bmleight: My original comment?  Implementation, I think.  I think the original code functions correctly (disables the button when it should), but I found myself needing to think...some when trying to follow the logic, so I threw out a potential simplification and general thoughts.

src/routes/forgot-password/components/ForgotPassword.js
rbkh: yep, great point! i think this was leftover from when i initially configured this to accept stuff from props and state on a different form (that I lazily copied and pasted!). So I'll update all the forms to be defined inline, yes?

src/components/Header/index.js
rbkh: how do you want me to author methods, bind in constructor? i'll update everywhere, just gimme the word!

src/routes/forgot-password/components/ForgotPassword.js
devinivy: Sounds good!

src/components/Header/index.js
devinivy: Until we collectively land on something else, I would suggest binding them in the constructor as we do in other projects.  (No need to bind methods that aren't passed directly as props to a component.)  We've  been exposed to different projects, though, so maybe I'm wrong in thinking that we don't do this elsewhere.  But maybe we should talk about this in strangeluv standup!

src/routes/login/components/Login.js
devinivy: I would suggest leveraging javascript to capture which field was blurred, passing it directly as an argument.
```js
onBlur={() => this.fieldBlurred('email')}
```

src/routes/forgot-password/components/ForgotPassword.js
devinivy: I don't think it's overkill :)

devinivy: There have been some forms that I skipped using strangeforms, and regretted it as the form grew.  But also, once you are familiar with its options, it saves some time even for very small forms.

blueprints/layout/files/__root__/layouts/__name__Layout/index.js
```diff
@@ -1,4 +1,5 @@
 const React = require('react');
+const T = require('prop-types');
```
devinivy: While this is a change I think we should make, I don't know if it belongs in the login recipe.

src/routes/forgot-password/components/ForgotPassword.js
devinivy: Ah!  Yeah, I guess I don't have a strong opinion!  This is more stylistic, but in the case of a function that returns a boolean based upon a single expression, I would probably just return that expression.
```js
    // When should the form be disabled?
    // When the field doesn't have a value or isn't an email!
    return !fieldHasValue || !IsEmail(email);
```

src/components/Header/index.js
bmleight: @jdonisvitch snuck it into cupie some, but that's the only one I'm aware of.

blueprints/layout/files/__root__/layouts/__name__Layout/index.js
rbkh: want me to make a separate pr or just leave it alone? I need to update the readme too, so i can do it around the same time

devinivy: Yeah, my suggestion would be to remove these changes from this PR, and possibly make a separate PR altogether.

src/selectors/auth.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+
+    getUserName: (state) => state.auth.credentials.user.firstName,
+    getIsAuthenticated: (state) => !!state.auth.isAuthenticated,
```
bmleight: I don't think the `!!` is doing anything

PR #220, opened 2018-05-11
Recipe initial load updates

wswoodruff:
The files full of the guts of this stuff are in the first commit, `Add utils` -- they're at the bottom of the diff so u can ignore 'em easily enough

- Most benefit is shown in the `actions/moon-phase` file
- Was able to confirm the `.FAILURE` part works by changing the url to a bogus one


src/action-types/index.js
```diff
@@ -0,0 +1,6 @@
+const { makeActionTypes } = require('utils/redux-helpers');
+
+module.exports = makeActionTypes({
+    COUNTER: require('./counter'),
+    MOON_PHASE: require('./moon-phase')
+});
```
wswoodruff: This file is a pain to me, keep forgetting to update it -- would be nice to think of some wiring to automate this

src/actions/moon-phase.js
```diff
@@ -1,69 +1,46 @@
...
-    return {
-        type: Types.MOON_LOAD_BEGIN
-    };
-};
+            // We'll want to lookup the moon phase for this very millisecond
+            const now = Date.now();
 
-// Marks the moment we've received some good info about the moon phase
-actions.loaded = (moon) => {
+            return new Promise((resolve, reject) => {
```
devinivy: I don't think this needs to be explicitly wrapped in a promise.

src/utils/redux-helpers/reducer-helpers.js
```diff
@@ -0,0 +1,30 @@
+// "Optimistic" in the sense that we'll update a value before we know
```
devinivy: This file probably doesn't need to stick around.

src/actions/moon-phase.js
wswoodruff: I wrapped this one because we're mutating the result before it goes to the reducers, so we can control when this action resolves

devinivy: If you are holding onto a promise (which you are, via `Axios.get()`) then you don't need to create one explicitly.  Returning `getMoonPhase` at the bottom of the promise body also isn't doing anything, I don't think.  This "passthrough" stuff seems leftover from thunk-land.

src/action-types/index.js
devinivy: Sounds good, let's think about that later, i.e. if we decide to go with this approach.

src/utils/redux-helpers/action-creator-helpers.js
```diff
@@ -0,0 +1,345 @@
...
+
+                async(request, { dispatch, getState })
+
+            ).then(
+                (rawResult) => {
+
+                    // Rethrow if Error
+                    if (rawResult instanceof Error) {
+                        throw rawResult;
+                    }
```
wswoodruff: Added these lines from the original source because wrapping the `async(...)` in `Promise.resolve()` seems to not cause an error to go to the second handler (line 150)

wswoodruff: This has been reverted -- I'm not sure what was going wrong when I wrote this -- it may be something with the hot module reloader

src/actions/counter.js
```diff
@@ -1,36 +1,33 @@
...
-    };
-};
+const T = require('prop-types');
+const { makeActionCreator } = require('utils/redux-helpers');
+const { COUNTER: Types } = require('action-types');
+
+const actions = exports;
+
+// Notice a limitation here -- can't set a default of 1 to increment
+actions.increment = makeActionCreator(
```
bmleight: What benefits does converting this into a simple action type give?  Consistency in how the app is written, and the ability to tie into `after` functionality?  I assume it'd be a pain to fire off the original increment action because of the changes to types?

Just thinking about what it might be like to upgrade an app to this style.  Would we need to hit every action in one swoop, or is there a reasonable way to port it over time?

src/utils/redux-helpers/selector-helpers.js
```diff
@@ -0,0 +1,43 @@
+const _ = require('lodash');
```
bmleight: This just came with the other redux helpers, right?

wswoodruff: Yes lodash is only visiting us because it's heavily baked-into the helpers

src/actions/counter.js
wswoodruff: > Why
Yeah just for consistency, to use both parts of these helpers -- `SIMPLE` and `ASYNC`

> Fire off the original increment action
Referring to the default setting of 1? If we want it there we can set it in initial state -- https://github.com/BigRoomStudios/strangeluv/pull/220/files#diff-ab974bcb2c9bcc1199dde0a37304b807R4

devinivy: > Just thinking about what it might be like to upgrade an app to this style. Would we need to hit every action in one swoop, or is there a reasonable way to port it over time?

Based upon my understanding, it could be done incrementally.  At the end of the day `actions.increment` is an action, either plain or a thunk.

src/utils/redux-helpers/selector-helpers.js
bmleight: Sorry, I meant this file

PR #222, opened 2018-05-18
exmple using redux-promise-middleware

bmleight:
Related to https://trello.com/c/CGj4EkLm/132-spec-out-data-fetching-strangeluv-component. 
 Example using https://github.com/pburtchaell/redux-promise-middleware

### DIFF
--- a/src/actions/moon-phase.js
+++ b/src/actions/moon-phase.js
@@ -3,67 +3,27 @@ const Types = require('../action-types/moon-phase');
 
 const actions = exports;
 
-// Marks the moment we start waiting for the moon phase
-actions.beginLoad = () => {
-
-    return {
-        type: Types.MOON_LOAD_BEGIN
-    };
-};
-
-// Marks the moment we've received some good info about the moon phase
-actions.loaded = (moon) => {
-
-    return {
-        type: Types.MOON_LOADED,
-        payload: moon
-    };
-};
-
-// Marks the moment an error occurred while look-up the moon phase
-actions.loadError = (error) => {
-
-    return {
-        type: Types.MOON_LOAD_ERROR,
-        payload: error
-    };
-};
-
-// We'll use a thunk to orchestrate the actions (listed above)
-// of this asynchronous process of hitting an API
 actions.load = () => {
 
-    // We'll want to lookup the moon phase for this very millisecond
     const now = Date.now();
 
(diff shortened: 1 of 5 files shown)


src/actions/moon-phase.js
```diff
@@ -3,67 +3,27 @@ const Types = require('../action-types/moon-phase');
...
-    return (dispatch) => {
-
-        // Indicate to the app that we're beginning to load the moon phase
-        dispatch(actions.beginLoad());
-
-        // Make a request to the Farmsense API
-        const getMoonPhase = Axios.get(`http://api.farmsense.net/v1/moonphases/?d=${now}`);
+    return {
+        type: 'MOON_LOAD',
+        payload: new Promise((resolve, reject) => {
```
devinivy: Would not suggest creating your own promise here—it's clunky to use the `resolve()` and`reject()` callbacks, and they aren't necessary here.

```js
payload: Axios.get(`http://api.farmsense.net/v1/moonphases/?d=${now}`).then((response) => {

    // Format of response from Farmsense API looks like [{ ... }]
    const result = response.data[0];

    // Bad result
    if (!result || result.Error) {
        throw new Error(response);
    }

    // Looks good!
    return result;
})
```

src/reducers/moon-phase.js
```diff
@@ -17,17 +15,17 @@ module.exports = (state, action) => {
...
             });
-        case MoonPhaseTypes.MOON_LOADED:
+        case 'MOON_LOAD_FULFILLED':
             return Object.assign({}, state, {
                 isLoading: false,
                 moonId: payload.Index,
                 error: null
             });
-        case MoonPhaseTypes.MOON_LOAD_ERROR:
+        case 'MOON_LOAD_REJECTED':
```
devinivy: It's interesting to me that they don't offer helpers for the action naming convention they use!

bmleight: In their ["complex" example](https://github.com/pburtchaell/redux-promise-middleware/blob/master/examples/complex/reducers/application/support.js#L9), they would do:

```
case `${MoonPhaseTypes.MOON_LOAD}_REJECTED`:
```

You can also pass in what the [suffixes should be](https://github.com/pburtchaell/redux-promise-middleware/blob/master/src/index.js#L21).

PR #223, opened 2018-05-21
taking out bootstrap classes that we're not using anymore

rbkh:
🙏 RIP bootstrap

### DIFF
--- a/src/layouts/CoreLayout/index.js
+++ b/src/layouts/CoreLayout/index.js
@@ -8,7 +8,7 @@ require('../../styles/core.scss');
 
 const CoreLayout = ({ children }) => (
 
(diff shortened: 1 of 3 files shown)


devinivy:
What about the line(s) that include bootstrap in the project?

rbkh:
That was already removed, I'm just cleaning up now-useless classes

devinivy:
Sweet, then I think the tests just need updating https://github.com/BigRoomStudios/strangeluv/blob/c460f6a90c61b9631d9bcaf3557293e777871e85/tests/components/Counter/Counter.spec.js

rbkh:
I actually just found that! cleaning up now.

PR #227, opened 2018-09-07
config tweak, prevent bunk login request on init

zemccartney:
yo! a few small dinks here.

- added a .env template file and set those vars in the main config, so you can configure the web client as you need

- made a hacky, ignorant tweak to the auth initializer. I made that change because on page load, if logged out, you'd see a login failed error, guessing b/c if local storage were available, the login action was fired off regardless of state of stuff in localStorage. I'm feeling 100% sure that this fucks some things up, shot it up here just to get ideas on paper. oh well, tried :trollface: 

otherwise, all's looking good I think. reset my password, got remembered and forgotten when I should have, signed in, logged in and logged out. thanks for all the work on this 🙏

### DIFF
--- a/.env-keep
+++ b/.env-keep
@@ -1,0 +1,8 @@
+# Rename me to .env then fill me with runtime configuration and credentials
+# Just don't try to check me into your repo :)
+# Confused? See https://github.com/motdotla/dotenv
+#
+# e.g.
+NODE_ENV=dev
+API_HOST=
+API_PREFIX=
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ node_modules
 
 dist
 coverage
+
+.env
--- a/config/main.js
+++ b/config/main.js
@@ -76,7 +76,9 @@ Edit at Your Own Risk
 // N.B.: globals added here must _also_ be added to .eslintrc
 config.globals = {
     'process.env': {
-        NODE_ENV: JSON.stringify(config.env)
+        NODE_ENV: JSON.stringify(config.env),
+        API_HOST: JSON.stringify(process.env.API_HOST),
+        API_PREFIX: JSON.stringify(process.env.API_PREFIX)
     },
     NODE_ENV: config.env,
     __DEV__: config.env === 'dev',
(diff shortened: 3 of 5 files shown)


rbkh:
@zemccartney I think those changes to the initializer introduces that auth redirect bug (where if you haven't had a login attempt yet and you hit an authenticated route, it doesn't redirect to `/login`). Firing this up seems to confirm that (plus I ran into a weird thing that I can't reproduce where it logged me in, and redirected to dashboard, but errors happened and nothing rendered). 

More context about this bit of code [here](https://github.com/BigRoomStudios/strangeluv/pull/219#discussion_r189063561)

So I think we may need a smarter way to manage errors!

bmleight:
I still have some issues with that piece of code.  Running `store.dispatch(AuthActions.login({ token: false }));` causes an unnecessary API request to be made, and in tern, an error that can be hard to tell whether or not we need to pass to the user or not.  We are silently ignoring it here, but that has been harder to do in the strangeluv native adaptation I've been working on.

rbkh:
good point @bmleight ! for the native adaptation, we don't _really_ care about authenticated wrappers on routes because we have better control over how and when a user navigates to different screens or not, right? If that's the case, maybe it's less important in the native auth recipe for the auth status to settle naturally.

devinivy:
> we don't really care about authenticated wrappers on routes because we have better control over how and when a user navigates to different screens or not, right?

I think we would defensively still want to handle authentication per route rather than every link in the app having to be correct about what pages the user is allowed to go to.  It also serves as a clear documentation (via code) of where the user is allowed to go in certain states.

zemccartney:
h'ok, so it turned out the redirect wasn't happening b/c `connectedRouterRedirect.authenticatingSelector` was enabled on both WAITING and INIT (auth statuses). I just removed INIT, so you'd see the unrendered Dashboard page only if you initiated a login request, then navigated to that page before the login finished

how does that change sound to y'all?

devinivy:
Sad to close, but this is stale now and the recipe is going to change quite a bit, including how we handle this issue.  Thank yalls 4 your service 🙏

PR #228, opened 2018-10-05
Recipe login material

tylerconstance:
This is ready for review. I picked up @rbkh's work and finished applying Material styles to the auth-related routes.

You can take a peek at the [Trello Card](https://trello.com/c/VTTXbh2P), but this is basically what's here:

- Shifted header nav items to the right and gave the Strangeluv label `overflow: ellipsis` to let it hide away on smaller screens, which is how this is handled in a few of the Material-UI [page layout demos](https://material-ui.com/page-layout-examples/pricing/). This looks fine down to iPhone SE size.
- Materialized Forgot Password and Reset Password routes, and refined Sign Up and Log In routes
- Fixed a front-end validation bug on Sign Up and Reset Password pages related to the Password and Confirm Password fields
- Sampled two colors from a gradient used in the BRS logo for theme colors. We can adjust or refine these as needed.

I'll make a few comments below where I had particular implementation questions.

Also, there's currently a server-side bug that will `null`-ify your password if you try to use the reset password flow, so you'll want to refrain from doing that.


src/components/Header/index.js
```diff
@@ -49,13 +69,36 @@ module.exports = class Header extends React.Component {
...
+            <AppBar
+                style={{ position: 'relative' }}
+                color='secondary'
+            >
+                <Toolbar>
+                    <Typography
+                        noWrap
+                        variant='title'
+                        color='inherit'
+                        style={{ flex: 1 }}
```
tylerconstance: There are a few instances in this file where I used inline styles for a one-style adjustment. It felt like overkill to create a styled-component for these, but I can definitely do that if it makes the most sense.

src/layouts/CoreLayout/styles.scss
```diff
@@ -1,3 +1,10 @@
 .mainContainer {
```
tylerconstance: We'll probably want to make this container a separate styled-component; I can refactor this before merging.

src/routes/login/components/Login.js
```diff
@@ -93,45 +95,78 @@ module.exports = class Login extends StrangeForms(React.Component) {
...
-                    onClick={this.submit}
-                    disabled={this.disableButton()}
-                >Login</button>
-            </form>
+            <FormWrapper>
+                <form onSubmit={this.submit}>
+                    <TextWrapper>
+                        <Typography variant='headline' gutterBottom>Log In</Typography>
+                        {this.props.errorMessage &&
+                            <FormHelperText>Oops, something went wrong! {this.props.errorMessage}</FormHelperText>
```
tylerconstance: It would be helpful to have a language style guide for messages and titles. Elsewhere this reads "Error: *message*," and it should be consistent throughout the site. I'm not a big fan of, "Oops," but I'm also not much fun in general.

wswoodruff: Yeah it'd be cool to have a constants file that has error prefixes depending on which type of error it is

src/layouts/CoreLayout/styles.scss
wswoodruff: If you see changes that should belong in the base `recipe-login` I think everyone would be fine with getting the changes in there then merge into here

wswoodruff: Or y'know depending on how far back the changes go, back to strangeluv vanilla

tylerconstance: That makes sense! Are we planning on merging `recipe-login-material` into anything, or is this PR just for illustrative purposes? I do like the idea of keeping a totally style-less `recipe-login` for non-Material solutions.

wswoodruff: Yeah I _think_ that's the plan -- we will most likely be starting with `recipe-login-material` on a new project but yeah keeping the material out of `recipe-login` I think is the goal

src/components/Header/index.js
rbkh: yeah, overall i kinda struggle on when and why to use styled-components over other means....sometimes it feels like overkill when it's a quick, one-time inline fix for a specific component. i also kinda miss just writing good ol' fashioned scss...

src/layouts/CoreLayout/styles.scss
rbkh: Oh good call, I forgot this class and stylesheet exists lol

This branch and `recipe-login` will live separately (I believe), and this PR is just for review and discussion.

src/index.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
     <head>
         <title>Strangeluv</title>
+        <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
```
tylerconstance: @rbkh and I talked about how we want to bring in typefaces, and while `npm` methods sound interesting, neither of us is super clear on *how* to do that while making sure we don't bring in too many weights or variants, so the classic `<link>` seems like the sensible solution for now.

src/routes/login/components/Login.js
rbkh: Yep that's my fault! The message formatting depends on my mood for the day....I changed this to "oops" while adding material cause it felt more "fun" and human, less developer-y; obviously didn't carry that over to the other components

src/index.html
rbkh: 👍

src/components/Header/index.js
wswoodruff: Poor scss is dead for us loLL

src/routes/login/components/Login.js
wswoodruff: Humans are fun 🤖⚡️👽

src/routes/forgot-password/components/ForgotPassword.js
```diff
@@ -70,30 +72,41 @@ module.exports = class ForgotPassword extends StrangeForms(React.Component) {
...
+                            fullWidth
+                            value={this.fieldValue('email')}
+                            onChange={this.proposeNew('email')}
+                            onBlur={this.fieldBlurred}
+                            error={this.showEmailError()}
+                        />
+                        {this.showEmailError() &&
+                            <FormHelperText>
+                                Please enter a valid email address.
+                            </FormHelperText>
```
tylerconstance: Should I give `<FormHelperText>` something like `color='error'` to make the error message red, or should I leave it the default grey color?

src/components/Header/index.js
rbkh: we're saying scss is dead, but we have a header stylesheet to manage `activeClassName` styles on nav items. Since we're also using inline styles on different header components, should we just toss that in the stylesheet so at least everything lives in one place? or do something different with active route styling and remove the header stylesheet? This is why i struggle...

src/routes/forgot-password/components/ForgotPassword.js
wswoodruff: Good question -- more conventions thoughts -- would be a good topic for a group of us to think about,

We could wrap the `FormHelperText` in a styled-component and pass an `error` property, which sets the color to red, or just set the `color='error'` on all error fields

tylerconstance: In the [docs](https://material-ui.com/api/form-helper-text/#formhelpertext) it looks like there's already a `bool` prop called `error` that turns these red; I can just do that!

wswoodruff: Cool

wswoodruff: I think having the convention that error messages are red seems sensible

tylerconstance: I'll make it so 👍

src/components/Header/index.js
wswoodruff: In the single `scss` file or near the component I mean

rbkh: I'd look in a scss file at the component level for stuff like that

wswoodruff:

PR #229, opened 2018-10-15
First pass at putting a User List on the Dashboard

tylerconstance:
Put a list of all the users on the Dashboard view. You can look at the details on the [Trello card](https://trello.com/c/R8Ntt6bc/160-add-a-user-list-page-to-the-material-auth-recipe).

I wouldn't say this is ready for merge, but I _would_ welcome feedback on code & proposed functionality. I gave each of the `UserListItem`s a `button` prop, so maybe they could link to a user detail page with management options, or maybe we should just put an action or two on the list item itself, like the option to delete a user.

edit: I'd also like to think about implementing a `max-height` for the `UserList` on the dashboard so the component stays relatively well-contained and doesn't force the entire dashboard view to have an arbitrarily large computed `height`.

### DIFF
--- a/src/routes/dashboard/components/Dashboard.js
+++ b/src/routes/dashboard/components/Dashboard.js
@@ -1,23 +1,48 @@
 const React = require('react');
 const T = require('prop-types');
-const { PageWrapper } = require('styles/global-components.js');
-const { Typography } = require('@material-ui/core');
+const { PageWrapper, UserListItem } = require('styles/global-components.js');
+const { Typography, List, ListItemText } = require('@material-ui/core');
 
 module.exports = class Dashboard extends React.Component {
 
     static propTypes = {
-        firstName: T.string.isRequired
+        firstName: T.string.isRequired,
+        getUsers: T.func.isRequired,
+        users: T.array.isRequired
     };
+
+    componentDidMount() {
+
+        this.props.getUsers();
+    }
 
     render() {
 
-        const { firstName } = this.props;
+        const { firstName, users } = this.props;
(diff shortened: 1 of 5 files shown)


src/styles/global-components.js
```diff
@@ -26,5 +27,11 @@ module.exports = {
         @media (min-width: 600px) {
             padding: 0 24px;
         }
+    `,
+
+    UserListItem: Styled(ListItem)`
+        width: 100%;
+        max-width: 360px;
+        background-color: ${Theme.palette.background.paper} !important;
```
tylerconstance: I'm interested to know if there's a way for me to omit the `!important` flag here. This style works as expected normally, but when the `ListItem` has the `button` prop set, the `background-color` is changed somewhere else with a higher specificity.

I do like the idea of using the `theme` color to make it easier to support dark themes and whatnot later. It seems like this may become increasingly relevant on the web in years to come now that the latest macOS update has started a conversation around future detection & implementation.

src/routes/dashboard/components/Dashboard.js
```diff
@@ -1,23 +1,48 @@
...
+
+                            <UserListItem
+                                key={user.id}
+                                dense
+                                button
+                            >
+                                <ListItemText
+                                    primary={user.fullName}
+                                    secondary={user.email}
+                                />
```
tylerconstance: We can also show `created` and `updated` dates if that's of interest. I opted to keep it simple, but we have options!

wswoodruff:

PR #230, opened 2018-10-15
Create settings page material

tylerconstance:
Early progress based on [this Trello card](https://trello.com/c/PY3F20oZ/158-add-settings-page-to-material-auth-recipe). Largely copied from `signup`, this is waiting for an API endpoint to tie into, and will need a little more refinement once that's ready. I _did_ add some basic front-end validation to only allow a submission if a new, unique password is entered `or` if a name or email address changes, but I haven't gone as far as putting that all together in an object to send off. We also need to find an appropriate place to put a link to this settings page—let me know what you think.

Maybe it even makes sense to decompose some of this into a separate component to share with the `signup` route? Suggestions welcome!

Edit: Forgot to mention, testing this is sort of an adventure because strangeluv automatically redirects to `/dashboard` when authenticated, so you'll need to hit `/settings`, end up on `/dashboard`, and then user the browser's back button to see the settings page.

### DIFF
--- a/src/routes/settings/components/Settings.js
+++ b/src/routes/settings/components/Settings.js
@@ -39,10 +39,7 @@ module.exports = class Settings extends StrangeForms(React.Component) {
         this.strangeForm({
             fields: ['firstName', 'lastName', 'email', 'currentPassword', 'password', 'confirmPassword'],
             get: (someProps, field) => this.state[field],
-            act: (field, value) => this.setState({ [field]: value }),
-            getFormValue: {
-                '*': this.getFormValue.bind(this)
-            }
+            act: (field, value) => this.setState({ [field]: value })
         });
 
         this.fieldBlurred = this._fieldBlurred.bind(this);
@@ -50,16 +47,6 @@ module.exports = class Settings extends StrangeForms(React.Component) {
         this.showPasswordError = this._showPasswordError.bind(this);
         this.disableButton = this._disableButton.bind(this);
         this.submit = this._submit.bind(this);
-    }
-
-    getCheckedValue(e) {
-
-        return e.target.checked;
-    }
-
-    getFormValue(e) {
-
-        return e.target.value || '';
     }
 
     _fieldBlurred(ev) {
@@ -85,22 +72,17 @@ module.exports = class Settings extends StrangeForms(React.Component) {
     _disableButton() {
 
         const { firstName, lastName, email, currentPassword, password, confirmPassword } = this.state;
-        // const fieldHasValue = (firstName && lastName && email && password) !== '';
         const fieldHasChanged =
-            (firstName !== this.props.userDetails.firstName)
-            || (lastName !== this.props.userDetails.lastName)
-            || (email !== this.props.userDetails.email);
+            (firstName !== this.props.userDetails.firstName) ||
+            (lastName !== this.props.userDetails.lastName) ||
+            (email !== this.props.userDetails.email);
         const passwordsMatch = password === confirmPassword;
-        const requestPasswordChange = (password !== '')
-            && passwordsMatch
-            && (currentPassword !== password)
-            && (currentPassword !== '');
+        const requestPasswordChange = (password !== '') &&
+            passwordsMatch &&
+            (currentPassword !== password) &&
+            (currentPassword !== '');
 
-        // if (fieldHasValue && IsEmail(email) && passwordsMatch) {
-        //     return false;
-        // }
-
-        // TODO: Make sure this OR statement doesn't mess things up if something is half-filled or something
+        // When hooking this up, we'll want to make sure we only send relevant changes
         if (fieldHasChanged || (requestPasswordChange)) {
             return false;
         }
@@ -120,7 +102,6 @@ module.exports = class Settings extends StrangeForms(React.Component) {
     render() {
 
         return (
-
             <FormWrapper>
                 <form onSubmit={this.submit}>
                     <TextWrapper>
--- a/src/routes/settings/components/Settings.js
+++ b/src/routes/settings/components/Settings.js
@@ -37,11 +37,10 @@ module.exports = class Settings extends StrangeForms(React.Component) {
         };
 
         this.strangeForm({
-            fields: ['firstName', 'lastName', 'email', 'currentPassword', 'password', 'confirmPassword', 'rememberMe'],
+            fields: ['firstName', 'lastName', 'email', 'currentPassword', 'password', 'confirmPassword'],
             get: (someProps, field) => this.state[field],
             act: (field, value) => this.setState({ [field]: value }),
             getFormValue: {
-                rememberMe: this.getCheckedValue.bind(this),
                 '*': this.getFormValue.bind(this)
             }
         });
@@ -101,7 +100,8 @@ module.exports = class Settings extends StrangeForms(React.Component) {
         //     return false;
         // }
 
-        if (fieldHasChanged || requestPasswordChange) {
+        // TODO: Make sure this OR statement doesn't mess things up if something is half-filled or something
+        if (fieldHasChanged || (requestPasswordChange)) {
             return false;
         }
 
@@ -110,10 +110,9 @@ module.exports = class Settings extends StrangeForms(React.Component) {
 
     _submit(ev) {
 
-        const { firstName, lastName, email, password, rememberMe } = this.state;
+        const { firstName, lastName, email, password } = this.state;
 
         this.props.onSubmit({ email, password, firstName, lastName  });
-        this.props.rememberAct(rememberMe);
 
         ev.preventDefault();
     }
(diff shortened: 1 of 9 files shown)


src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+        return this.state.isBlurred.confirmPassword && (this.state.password !== this.state.confirmPassword);
+    }
+
+    _disableButton() {
+
+        const { firstName, lastName, email, currentPassword, password, confirmPassword } = this.state;
+        // const fieldHasValue = (firstName && lastName && email && password) !== '';
+        const fieldHasChanged =
+            (firstName !== this.props.userDetails.firstName)
+            || (lastName !== this.props.userDetails.lastName)
```
devinivy: It's a total nit (or not even, since it's purely arbitrary) but you'll typically see BRS code have multi-line conditions written like this,
```js
        const fieldHasChanged = (firstName !== this.props.userDetails.firstName) ||
            (lastName !== this.props.userDetails.lastName) ||
            (email !== this.props.userDetails.email);
```

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+            || (email !== this.props.userDetails.email);
+        const passwordsMatch = password === confirmPassword;
+        const requestPasswordChange = (password !== '')
+            && passwordsMatch
+            && (currentPassword !== password)
+            && (currentPassword !== '');
+
+        // if (fieldHasValue && IsEmail(email) && passwordsMatch) {
```
devinivy: Dead code like this can be nixed.

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+
+        this.props.onSubmit({ email, password, firstName, lastName  });
+
+        ev.preventDefault();
+    }
+
+    render() {
+
+        return (
+
```
devinivy: I've started seeing this empty line in our components, and I don't know it's origin!

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+                            variant='outlined'
+                            label='First Name'
+                            margin='normal'
+                            fullWidth
+                            value={this.fieldValue('firstName')}
+                            onChange={this.proposeNew('firstName')}
+                            onBlur={this.fieldBlurred}
+                        />
+                        <TextField
+                            id='lastName'
```
devinivy: Does this introduce an actual `id` attribute?  If so, I might suggest going with something else like `name`, or passing the name of the field directly to `fieldBlurred()`.

src/routes/settings/containers/Settings.js
```diff
@@ -0,0 +1,23 @@
...
+        errorMessage: state.auth.error.message,
+        userDetails: {
+            firstName: AuthSelectors.getUserName(state),
+            lastName: AuthSelectors.getUserLastName(state),
+            email: AuthSelectors.getUserEmail(state)
+        }
+    }),
+    {
+        // onSubmit: AuthAct.registerUser
+        onSubmit: (console.log('Trying to submit on Settings page'))
```
devinivy: This looks like it will only run once when the app initializes.  Do you want `() => console.log(/* ... */)` ?

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+        });
+
+        this.fieldBlurred = this._fieldBlurred.bind(this);
+        this.showEmailError = this._showEmailError.bind(this);
+        this.showPasswordError = this._showPasswordError.bind(this);
+        this.disableButton = this._disableButton.bind(this);
+        this.submit = this._submit.bind(this);
+    }
+
+    getCheckedValue(e) {
```
devinivy: This seems unused in here.

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+
+        return e.target.checked;
+    }
+
+    getFormValue(e) {
+
+        return e.target.value || '';
```
devinivy: If the default to `''` isn't necessary, then this doesn't need to be configured at all.  I also try to keep methods that don't need `this` either static (`static getFormValue(e)`) or in a helper (e.g. `internals.getFormValue = (e) => ...`).

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,235 @@
...
+        this.strangeForm({
+            fields: ['firstName', 'lastName', 'email', 'currentPassword', 'password', 'confirmPassword'],
+            get: (someProps, field) => this.state[field],
+            act: (field, value) => this.setState({ [field]: value }),
+            getFormValue: {
+                '*': this.getFormValue.bind(this)
+            }
+        });
+
+        this.fieldBlurred = this._fieldBlurred.bind(this);
```
devinivy: If all these `bind()`s hurt you, I think we (and the greater community) have become more open to starting with arrows,
```js
    onBlur={() => this.fieldBlurred()}
```

Either is cool though!

devinivy (commented):
Looks pretty good to me!  Left some notes.

src/routes/settings/components/Settings.js
tylerconstance: Oh, weird! I can change it to match the BRS convention if you prefer, but I'm partial to seeing the operation at the start of the line. To me, the difference is like

>123 x
4

compared to

>123
x 4

src/routes/settings/components/Settings.js
tylerconstance: Good question! I pulled this directly from `Signup.js`, but I can do some testing to see if it works fine sans configuration as a static or helper function.

src/routes/settings/components/Settings.js
devinivy: I think it's alright to have both.  It's new-ish that this is allowed by our react linting rules, so I expect we'll be used to seeing/understanding both.

src/routes/settings/components/Settings.js
devinivy: It was propagated from the strange-forms example, which probably originated from me!  I'll make a point to change it.  https://github.com/BigRoomStudios/strange-forms#example

src/routes/settings/components/Settings.js
devinivy: Yes, please!  We go by the hapi style guide, which has a rule about this: https://hapijs.com/styleguide#operators

Maybe it goes without saying, but some things that look weird to one person wont look weird to another.  Both of those examples above feel equally ergonomic to me.  A few people here felt really weird moving to [Stroustrup curlies](https://en.wikipedia.org/wiki/Indentation_style#Variant:_Stroustrup), for example, but over time it felt natural.

tylerconstance: I'll make it so!

And totally—I got over using four spaces instead of two, so this is no big deal :)

src/routes/settings/components/Settings.js
tylerconstance: What a fun discovery! Do you have an implementation preference between making it `static` vs attaching it to `internals`? We may as well pick your favorite! :)

devinivy: :)

I think my preference would actually be to inline it, come to think of it!  If it's shared across multiple forms, perhaps it would just end-up in a utility.

src/routes/settings/components/Settings.js
tylerconstance: According to the [material docs](https://material-ui.com/api/text-field/#props),

>The id of the input element. Use that property to make label and helperText accessible for screen readers.

I'm not sure if that prop is doing something for me automatically or not, though.

devinivy: Got it!  Yes, carry on! :P

src/routes/settings/components/Settings.js
tylerconstance: Cool! Right now it looks like it's on `ForgotPassword`, `Login`, `ResetPassword`, `Settings`, and `SignUp`, so maybe a utility makes sense, but at the same time, it's just an itty bitty one-liner. What are your thoughts here?

devinivy: Not a big deal to me either way.  The main thing on my mind is: do any of them need it?

tylerconstance: ^ We chatted about this in-person, but the short answer is "no."

`(e) => e.target.value || ''` is already the default behavior, so in simple cases, `getFormValue()` can be omitted altogether. In other cases, like for a checkbox, we can do something like this:

```js
getFormValue: {
    rememberMe: (e) => e.target.checked,
    '*': (e) => e.target.value || ''
}
```

I thought maybe we could omit the `'*': (e) => e.target.value || ''` case altogether, but that didn't work for me, and it's not bothering me to keep it.

devinivy: Ah, yup, I see in the source it applies the default a little too eagerly and misses the configured-per-field case 😬

tylerconstance: Should we create an issue in [strange-forms](https://github.com/BigRoomStudios/strange-forms)?

devinivy: Sure!

wswoodruff:

PR #231, opened 2018-10-16
 Fix deep-linking when authenticated

devinivy:
The login action invariably caused the user to be directed to the `/dashboard`.  Now the login action performs no redirection, and the caller can decide whether or not to do that on their own.  I also fixed a minor issue where settling login state required a network request even if no token was remembered.

### DIFF
--- a/src/reducers/auth.js
+++ b/src/reducers/auth.js
@@ -46,6 +46,10 @@ module.exports = (state, action) => {
 
         case StrangeAuth.types.LOGIN_FAIL:
 
+            if (payload.code === 'NO_TOKEN_ON_INIT') {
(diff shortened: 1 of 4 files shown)


src/reducers/auth.js
```diff
@@ -46,6 +46,10 @@ module.exports = (state, action) => {
 
         case StrangeAuth.types.LOGIN_FAIL:
 
+            if (payload.code === 'NO_TOKEN_ON_INIT') {
+                return state;
+            }
```
wswoodruff: Did this always error if the user started the app and attempted to login with no stored token? Thanks for this error code 👍

PR #232, opened 2019-05-12
Update middle end

wswoodruff:
- Adapt middle-end from our recent best practices
- Update existing counter example to use new middle-end
- Aims to give examples of most of the features of the middle-end


wswoodruff:
Stay tuned for a new PR with the new https://github.com/bigroomstudios/strange-middle-end

PR #233, opened 2019-09-14
(WIP) Give a good start to moving strange-middle-end to strangeluv

wswoodruff:
- Replace wiring and old setup for actions, action-types, reducers, selectors, initializers with usage of helpers from `strange-middle-end`


src/middle-end/index.js
```diff
@@ -0,0 +1,33 @@
+const Redux = require('redux');
+const { routerMiddleware: RouterMiddleware } = require('react-router-redux');
+const { middleware: { thunk, errorLogger }, ...M } = require('strange-middle-end');
```
devinivy: To fix your build issues, try using this instead for now:
```suggestion
const { middleware: { thunk, errorLogger }, ...M } = require('strange-middle-end/dist/strange-middle-end');
```

It looks like the ES module version has some issues with the peer dependencies.  Gotta look into what rollup is doing there.  The commonjs version seems to be working fine.  I also need to checkout why the sourcemaps aren't being used!

devinivy: P.S. this is good for @zemccartney to know, since he's working-out some of these kinks right now by integrating the middle-end into a project.

wswoodruff: Cool will try thanks!

devinivy: I recommend using a different variable here.  Maybe just `MiddleEnd`, in order to distinguish require('strange-middle-end')` from `require('./middle-end')` in the project, which are two different things.

src/middle-end/mods/counter/action-types.js
```diff
@@ -0,0 +1,6 @@
+
+const M = require('middle-end');
```
devinivy: I think you'll find that you want to use strange-middle-end here.

src/middle-end/mods/counter/actions.js
```diff
@@ -0,0 +1,8 @@
+
+const M = require('middle-end');
```
devinivy: Ditto: I think you'll find that you want to use strange-middle-end.

src/middle-end/mods/counter/reducer.js
```diff
@@ -0,0 +1,12 @@
+
+const M = require('middle-end');
```
devinivy: Again here :)

src/middle-end/index.js
```diff
@@ -0,0 +1,33 @@
...
+    // which creates a circular dependency since mods are needed here in 'M.create'.
+    // Specifying 'mods' here as a function allows our mods to require this middle-end
+    // lazily and in a controlled way, which is at some point after 'M.initialize()' is called.
+    // In this app, 'M.initialize()' is called in /src/main.js.
+    mods: () => require('./mods'),
+    createStore: (reducer) => {
+
+        return Redux.createStore(
+            reducer,
+            {},
```
devinivy: If this is not used, I believe it can be omitted.
```suggestion
```

src/middle-end/index.js
zemccartney: mmm! very cool, thanks for hashing this out and the heads-up 🙏

devinivy: This should be taken care of in v0.1.1 https://github.com/rollup/rollup-plugin-commonjs/issues/410

wswoodruff:
This branch will now compile and start up! Routing to `counter` errors on something so that'll have to be fixed

src/middle-end/index.js
wswoodruff: All settled now with the new version and the require statements not digging into `/dist`

devinivy:
Thanks for this @wswoodruff!  Now incorporated into #234

PR #234, opened 2019-09-30
Version 2

devinivy:
- Update to depend on node v10/12.
- Update to webpack v4.
- Reorganize webpack configuration.
- Update babel deps to use new loaders and @babel-scoped packages.
- Update eslint and related plugin/config deps.
- Remove npm-better-run.
- Remove bin/ folder (if practical under webpack v4).
- Remove layouts/ folder.
- Remove redux-cli and blueprints.
- Remove nodemon.
- Replace mocha/chai/karma/sinon/enzyme testing suite with jest and @testing-library/react.
- Incorporate styled-components as the default CSS solution, removing previous config/deps related to CSS.
- Incorporate strange-middle-end and remove old middle-end deps such as keymirror.
- Move from react-router-redux to connected-react-router.
- Decouple router parent component from rendered routes.
- Default 404 and error boundary.
- Update docs.


src/App.js
```diff
@@ -0,0 +1,31 @@
...
+const StrangeRouter = require('strange-router');
+const { default: ErrorBoundary } = require('react-error-boundary');
+const ErrorFallback = require('./components/ErrorFallback');
+const GlobalStyle = require('./components/GlobalStyle');
+const Routes = require('./routes');
+
+module.exports = ({ store, Router = ConnectedRouter, ...routerProps }) => {
+
+    return (
+        <ErrorBoundary FallbackComponent={ErrorFallback}>
```
jdonisvitch, line 15: Very nice!

src/components/ErrorFallback.js
```diff
@@ -0,0 +1,16 @@
+const React = require('react');
+
+module.exports = () => {
+
+    return (
+        <>
+            <h3>Oops! Something went wrong.</h3>
+            <h4>
+                You can keep browsing or try{' '}
```
jdonisvitch, line 9: `try{' '}` looks odd?

jdonisvitch: pardon if it's just something I'm unaware of, heh!

config/eslint.config.js
```diff
@@ -0,0 +1,35 @@
...
+    parser: 'babel-eslint',
+    env: {
+        browser: true,
+        commonjs: true,
+        es6: true,
+        jest: true,
+        node: true
+    },
+    plugins: [
+        'jsx-a11y',
```
jdonisvitch, line 13: I just stumbled into this as well, hadn't used it yet. 🤔 👍

package.json
```diff
@@ -1,157 +1,81 @@
 {
   "name": "strangeluv",
-  "version": "0.0.0",
-  "description": "How I Learned to Stop Worrying and Love React",
+  "version": "2.0.0",
+  "description": "My webpack project",
```
jdonisvitch: Are we gonna change the desc?

src/routes/index.js
```diff
@@ -1,14 +1,27 @@
...
-// Create routes
-module.exports = (store) => ([{
-    path: '/',
-    component: CoreLayout,
-    childRoutes: [
-        Home,
-        CounterRoute(store)
-    ]
-}]);
+const CounterPage = React.lazy(() => import('./counter/containers/CounterPage'));
```
jdonisvitch, line 7: 💯

jdonisvitch (approved):

src/components/ErrorFallback.js
devinivy: I need to add a space between `try` and the `<button />` element on the next line.  The way JSX whitespace works, this seemed like the most direct way to do it to me, but I agree it's kinda awkward.

package.json
devinivy: Good catch!  I'm thinking it might make sense to clear it out altogether.

src/components/Header.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = () => {
+
+    const { Link } = internals;
+
+    return (
+        <div>
+            <h1>Strangeluv</h1>
+            <Link exact to='/'>Home</Link>
+            {' · '}
```
tylerconstance: This line causes a "Failed to compile" error for me:

>Line 15:13:  Curly braces are unnecessary here  react/jsx-curly-brace-presence

devinivy: I think the behavior might have changed when they fixed this issue: https://github.com/yannickcr/eslint-plugin-react/issues/2423 . Hmm...

If they are unnecessary, do you mind removing them?  If they are necessary and this is a false positive, you can downgrade eslint-plugin-react prior to the bug was introduced.

devinivy: It also might have an autofix worth trying!

tylerconstance: It looks like it's not valid because of the period.

```
 * there's no way to inject a whitespace into jsx without a container so this
 * will always be allowed.
 */
<App>{' '}</App>
<App>{'     '}</App>
```
([documentation](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md))
So I think you could technically do {' '}.{' '}?

devinivy: It's probably this bug that has a fix that hasn't been released yet: https://github.com/yannickcr/eslint-plugin-react/issues/2427

When this PR went-up it definitely compiled fine and passed the CI checks.  In the past couple days eslint-plugin-react has made several changes to this particular lint rule, so I expect they are working out some kinks.  I would recommend downgrading locally for now until things settle there.  Try `npm install --save-dev eslint-plugin-react@7.14.3`.

tylerconstance: Downgrading `eslint-plugin-react` didn't actually do it, but downgrading `eslint` did the trick for now.

devinivy: Huh, pretty odd!  Glad you worked around it for now.

config/eslint.config.js
wswoodruff: If anyone else is wondering, `a11y` stands for `accessibility`. The number in the middle is how many letters are in there. In the same vein `i18n` stands for `internationalization`

package.json
wswoodruff: Gonna go ahead and remove this

src/components/Header.js
wswoodruff: I ran into this same issue, even after upgrading to `v7.16.0` which was _supposed_ to fix it (This PR https://github.com/yannickcr/eslint-plugin-react/pull/2431)

Someone else commented that it actually didn't fix the issue and they opened a new one. Last comment was 3 hours ago https://github.com/yannickcr/eslint-plugin-react/issues/2454)

wswoodruff: I don't want to spend too much time on this. I put in a temp fix and comment about it by declaring a var for this
```
const divider = ' · ';
...
{divider}
```

wswoodruff: This part of the code is changed in Tyler's material branch anyway so we'll be fine

package-lock.json
```diff
@@ -0,0 +1,12106 @@
+{
```
devinivy: I was hoping to omit a package-lock from the repository, if that's cool!  I still think projects should utilize one, but I also think it makes sense to have a completely fresh installation with the latest patch versions of everything upon starting a new project.

wswoodruff: Ohp! My bad! Didn't realize that and that makes sense to me

src/components/ErrorFallback.js
wswoodruff: Strange how the braces issue doesn't complain about this but complains about the other, /shrug

PR #235, opened 2019-10-04
V2 mui

tylerconstance:
I think this is ready for feedback. There may still be some weirdness with `eslint` or `eslint-plugin-react` since I didn't make any changes to the `package.json`. I want to hold out a little longer to see if an update fixes the issue with `{' '}` causing a compile error.


wswoodruff (approved):
Noice!

🎨👨‍🎨

PR #236, opened 2019-10-14
Removed material-ui from project

tylerconstance:
Creating a draft PR for code review and comments on this recipe.
This removes material-ui and several styles from the project to get closer to an unstyled version.

~There are a few open questions I want to answer before I consider this good to go:~
- ~Should we include a default or commented-out `theme`?~ We added a theme with an example color object
- ~Should we re-introduce the GlobalStyles component?~ Re-introduced normalize styles and an example of a `body` style
- ~Should we remove _all_ css from this branch, or keep some in as a means of example usage?~ Kept a few examples
- ~Should I remove the `package-lock` from this branch?~ Poof💨, gone


devinivy:
Here are my opinions on the matter.

 - Should we include a default or commented-out theme?

   I think we should include an empty theme for styled components, but not a MUI theme.

 - Should we re-introduce the GlobalStyles component?

   I would recommend using the [`normalize()`](https://polished.js.org/docs/#normalize) styles from polished.  After some research, that's how I left things before introducing MUI.

 - Should we remove all css from this branch, or keep some in as a means of example usage?

   I would recommend keeping some examples of styled components.

 - Should I remove the package-lock from this branch?

   I would recommend that we do remove it.  We don't include it on the main branch, for one. 
 But also, here was my thought last time it came up https://github.com/BigRoomStudios/strangeluv/pull/234#discussion_r334117009:
   > I was hoping to omit a package-lock from the repository, if that's cool! I still think projects should utilize one, but I also think it makes sense to have a completely fresh installation with the latest patch versions of everything upon starting a new project.

tylerconstance:
> I think we should include an empty theme for styled components, but not a MUI theme.

Should we provide any structure to what that theme should look like? And should it be a separate file, or just a variable or something on `internals` in this file?

devinivy:
IMO it doesn't need any structure, but it should be in a separate file.  What do you think?

src/theme.js
```diff
@@ -1,3 +1,7 @@
-const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiTheme');
+const theme = {
+    colors: {
+        duckBill: '#DB2F2A'
```
tylerconstance: I may swap this for a more generic name like "primary" or "secondary," similar to how we did this with the material-ui theme in [the material branch](https://github.com/BigRoomStudios/strangeluv/pull/237#discussion_r336609394)

bmleight (approved):

tylerconstance:
Closing since the purpose of this draft PR wasn't to merge into a branch, but to be reviewed as a standalone branch.

PR #237, opened 2019-10-18
Passed theme into App component and showed example usage

tylerconstance:
Adds a few things that were missing from the v2-mui branch, like passing the theme component through props to the App component in `index.js` and showing a few examples with the mui-theme and styled-components.

### DIFF
--- a/src/routes/home/components/HomePage.js
+++ b/src/routes/home/components/HomePage.js
@@ -7,11 +7,11 @@ const internals = {};
 
 module.exports = () => {
 
-    const { Image, HomepageContainer } = internals;
+    const { Image, HomepageContainer, WelcomeMessage } = internals;
 
     return (
         <HomepageContainer>
-            <Typography variant='h4' align='center'>Welcome!</Typography>
+            <WelcomeMessage variant='h4' align='center'>Welcome!</WelcomeMessage>
             <Image
                 alt='This is a duck, because Redux!'
                 src={DuckImage}
@@ -29,4 +29,10 @@ internals.Image = Styled.img`
 internals.HomepageContainer = Styled.div`
     align-self: center;
     margin: auto;
+`;
+
+internals.WelcomeMessage = Styled(Typography).attrs({ variant: 'h4', align: 'center' })`
+
+    // Example leveraging the mui theme from styled-components
+    color: ${(props) => props.theme.palette.duckYellow.main};
 `;
--- a/src/routes/home/components/HomePage.js
+++ b/src/routes/home/components/HomePage.js
@@ -34,5 +34,5 @@ internals.HomepageContainer = Styled.div`
 internals.WelcomeMessage = Styled(Typography).attrs({ variant: 'h4', align: 'center' })`
 
     // Example leveraging the mui theme from inside a styled-component
-    color: ${(props) => props.theme.palette.duckYellow.main};
+    color: ${({ theme }) => theme.palette.duckYellow.main};
 `;
--- a/src/routes/home/components/HomePage.js
+++ b/src/routes/home/components/HomePage.js
@@ -11,7 +11,7 @@ module.exports = () => {
 
     return (
         <HomepageContainer>
-            <WelcomeMessage variant='h4' align='center'>Welcome!</WelcomeMessage>
+            <WelcomeMessage>Welcome!</WelcomeMessage>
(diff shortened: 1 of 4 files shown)


src/routes/home/components/HomePage.js
```diff
@@ -30,3 +30,9 @@ internals.HomepageContainer = Styled.div`
     align-self: center;
     margin: auto;
 `;
+
+internals.WelcomeMessage = Styled(Typography).attrs({ variant: 'h4', align: 'center' })`
+
+    // Example leveraging the mui theme from inside a styled-component
+    color: ${(props) => props.theme.palette.duckYellow.main};
```
devinivy: It's purely stylistic, but this makes me personally feel warm/fuzzy.

```suggestion
    color: ${({ theme }) => theme.palette.duckYellow.main};
```

devinivy: Does this cause tests to fail?  I could imagine `duckYellow.main` not existing on the default theme in a way that would cause a hard failure.

src/theme.js
```diff
@@ -1,3 +1,12 @@
 const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiTheme');
 
-module.exports = CreateMuiTheme({});
+module.exports = CreateMuiTheme({
+    palette: {
+        duckYellow: {
```
devinivy: I wonder if the example of theming should be something more likely to not need changing/removing when starting a project.  Otherwise all of our projects will inevitably end-up with `duckYellow` in the theme :P

tylerconstance: Yeah, that sounds good! Do you have any suggestions?

src/routes/home/components/HomePage.js
tylerconstance: Yeah, this is way better!

src/theme.js
devinivy: Here are two thoughts:

1. Pull the color from MUI: https://material-ui.com/customization/color/#color
2. Make the secondary text color yellow for the application, rather than adding a new branch of the theme.  This will organically be changed for each application.  Would also address the testing issue from the other comment, if that's a real issue.

tylerconstance: These are some strong ideas, and I decided to implement both to show usage https://github.com/BigRoomStudios/strangeluv/pull/237/commits/1a54f9cfc90fd4672c742e31f0fbbea2e0762c1f

I'm using the `secondary` color in the `HomePage` component, and pulled in `amber` to use `amber[500]` as the secondary color.

src/theme.js
```diff
@@ -1,3 +1,13 @@
 const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiTheme');
+const { default: amber } = require('@material-ui/core/colors/amber');
```
tylerconstance, line 8: _Slightly_ abusing the GitHub suggestion feature to add a comment that @wswoodruff mentioned would be beneficial. It seems apt to include that in this PR.

```suggestion

// The object below overrides specific values and/or extends the default material-ui theme, which can currently be found here: https://material-ui.com/customization/default-theme/#default-theme

// Any of the values in that default theme may be used throughout this project, even if they aren't explicitly defined here (e.g., `theme.shape.borderRadius` will return 4, even though `shape` isn't currently defined in this file)

```

src/theme.js
```diff
@@ -1,3 +1,13 @@
...
```
devinivy: Just our (via hapi) convention to `require()` things in pascal-case:

```suggestion
const { default: Amber } = require('@material-ui/core/colors/amber');
```

src/theme.js
```diff
@@ -1,3 +1,13 @@
...
 
-module.exports = CreateMuiTheme({});
+module.exports = CreateMuiTheme({
+    palette: {
+        secondary: {
+            main: amber[500]
```
devinivy: I believe you can just set `secondary: Amber` and that will populate all the secondary colors based on amber, with main being `Amber[500]` as you have set here.

devinivy: More info: https://material-ui.com/customization/palette/

tylerconstance: Ooh, yeah, that's really cool. I made the change!

devinivy (approved):

bmleight (approved):

PR #239, opened 2019-11-20
Set up a create account page and handled account creation

tylerconstance:
A reimagining of https://github.com/BigRoomStudios/strangeluv/pull/219.

I set up a back end with `user-pal` and then wrote the front and middle ends to handle account creation. There's currently no fanciness to automatically login or redirect, and no messaging, either.

We can implement material-ui's `snackbar` for the latter, or we can come up with another solution.

This work just handles account creation, but there are a few other features to add before this recipe is ready to go:

- Implement Login & Logout
- Remember login
- Implement "Forgot password" flow
- Create an authenticated route
- Adding niceness like redirecting and automatically logging in

There are also a few things I'd like to take care of for this account creation feature work:

- Clean up some comments and commented code
- Prevent submission if any fields are missing
- Add some kind of messaging to communicate success and errors (feedback welcome on how to approach this)
- Remove/replace inline styles

I'm not sure how we should approach development on this recipe, but maybe we can focus on getting _this_ code in a good place to act as a decent "base," then make PRs for additional features to this branch, until eventually we can call this recipe done? Open to suggestions!


src/utils/web-client.js
```diff
@@ -0,0 +1,30 @@
+const Axios = require('axios');
+
+const client = module.exports = Axios.create({
+    baseURL: process.env.API_BASE_URL,
+    responseType: 'json',
+    headers: { common: {} },
+    transformResponse: (data) => {
```
devinivy: I believe this `transformResponse` business can be removed.  It looks like it's catered to some other API than user-pal.

tylerconstance: 1000% copy and pasted this file from another project 🍝

src/routes/join/components/SignupPage.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    handleSubmit = async (ev) => {
+
+        ev.preventDefault();
+        const accountInfo = this.formatFields();
+        this.setState({ isSubmitting: true });
+        const [err, result] = await this.props.reqCreateAccount(accountInfo);
+        this.setState({ isSubmitting: false });
+        if (!err && result.success) {
```
devinivy: I think this `result.success` stuff can go, since it looks like it catered to some other API than user-pal.
```suggestion
        if (!err) {
```

src/routes/join/components/SignupPage.js
```diff
@@ -0,0 +1,180 @@
...
+    };
+
+    static fields = {
+        firstName: '',
+        lastName: '',
+        email: '',
+        password: ''
+    };
+
+    defaultSubmit = (ev) => {
```
devinivy: I personally find it nice to keep methods below class constructors, not sure what others think!

src/middle-end/auth/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    // reducer: require('./reducer'),
```
devinivy:
```suggestion
```

src/middle-end/auth/index.js
```diff
@@ -0,0 +1,5 @@
...
+    actions: require('./actions')
+    // selectors: require('./selectors')
```
devinivy:
```suggestion
```

src/routes/join/components/SignupPage.js
tylerconstance: should we cut [result] from the destructuring too?

devinivy: If you don't need it then yeah, I'd say so!  I think the linter might even give you a little warning about the unused variable.

src/routes/join/components/SignupPage.js
devinivy: Actually, looks like this isn't used anyway :)

src/routes/join/components/SignupPage.js
```diff
@@ -0,0 +1,180 @@
...
+    render() {
+
+        const { isSubmitting, showPassword } = this.state;
+        const { HomepageContainer } = internals;
+
+        return (
+            <HomepageContainer>
+                <Typography variant='h4' align='center' gutterBottom>Sign Up</Typography>
+                <form
+                    style={{ display: 'flex', flexDirection: 'column' }}
```
devinivy: I don't have strong feelings, but this would be a place where we could use box as a drop-in replacement for inline styles: `<Box component='form' display='flex' flexDirection='column' ...`.

tylerconstance: Oh snap, all these inline styles were meant to be temporary. I agree with this!

devinivy:
@tylerconstance might be worth taking a peek at why the CI build is failing.

src/utils/web-client.js
devinivy: 🧀

src/routes/join/components/SignupPage.js
```diff
@@ -0,0 +1,180 @@
...
+            </HomepageContainer>
+        );
+    }
+};
+
+internals.Image = Styled.img`
+    display: block;
+    width: 120px;
+    margin: 1.5rem auto;
+`;
```
tylerconstance: I need to cut this, too.

src/routes/join/components/SignupPage.js
```diff
@@ -140,7 +134,6 @@ module.exports = class SignupPage extends StrangeForms(React.Component) {
                                     <IconButton
                                         aria-label='toggle password visibility'
                                         onClick={this.handleClickShowPassword}
-                                        onMouseDown={this.handleMouseDownPassword}
```
devinivy: Isn't something like this needed to avoid losing focus on the input?

tylerconstance: Oh, I think you might be right. I'll fix this.

devinivy:
I propose we close this out once #249 is merged into here and the tests appear passing.

devinivy:
#249 is merged and the tests are passing, so I think this one is good to go 🎉

devinivy:
Don't delete the branch 👇

PR #240, opened 2019-11-25
Create login view

tylerconstance:
Creates a `/login` view and allows a user to log in and log out.

The test is currently failing because the middle-end used in the test is different than the middle-end used in `containers/Layout.js`. Devin says this is a tricky pain-in-the-butt problem, so I'm not going to worry about that right now.

As a placeholder, when the user is authenticated, a logout button appears in the header. We can handle this any number of ways, though:

- Add a login button in the header when the user is not authenticated
- Add login/signup buttons to the homepage
- Add login/signup fields to the homepage
- Add a ︙ menu to the header with a log in or log out option beneath it
- Add a logout icon in lieu of a logout button

There's also a line that can be changed after a new feature is implemented in `strange-middle-end`.

### DIFF
--- a/src/routes/login/components/LoginPage.js
+++ b/src/routes/login/components/LoginPage.js
@@ -1,0 +1,120 @@
+const React = require('react');
+const T = require('prop-types');
+const { default: Styled } = require('styled-components');
+const { default: Typography } = require('@material-ui/core/Typography');
+const { default: TextField } = require('@material-ui/core/TextField');
+const { default: Button } = require('@material-ui/core/Button');
+const { default: Box } = require('@material-ui/core/Box');
+const { default: Link } = require('@material-ui/core/Link');
+const StrangeForms = require('strange-forms');
+
+const internals = {};
+
+module.exports = class LoginPage extends StrangeForms(React.Component) {
+
+    static propTypes = {
+        reqCreateAccount: T.func.isRequired
+    };
+
+    static fields = {
+        email: '',
+        password: ''
+    };
+
+    constructor(props) {
+
+        super(props);
+
+        this.state = {
+            email: '',
+            password: '',
+            isSubmitting: false
+        };
+
+        this.strangeForm({
+            fields: Object.keys(LoginPage.fields),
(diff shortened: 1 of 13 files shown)


src/routes/login/components/LoginPage.js
```diff
@@ -0,0 +1,121 @@
...
+
+        this.state = {
+            email: '',
+            password: '',
+            isSubmitting: false
+        };
+
+        this.strangeForm({
+            fields: Object.keys(LoginPage.fields),
+            get: (someProps, field) => {
```
devinivy: This looks like it's trying to nab default/controlled values off of a prop named `data`.

src/utils/web-client.js
```diff
@@ -6,9 +6,9 @@ const client = module.exports = Axios.create({
...
 
-client.updateConfiguration = ({ apiBaseUrl, ...rest }) => {
+client.updateAuth = ({ token }) => {
 
-    if (apiBaseUrl || apiBaseUrl === '') {
-        client.defaults.baseURL = apiBaseUrl;
-    }
+    const headers = client.defaults.headers.common;
+
+    return Object.assign(headers, { authorization: `Bearer ${token}` });
```
devinivy, line 13: I don't think this is being cleared on logout yet.

src/routes/login/components/LoginPage.js
tylerconstance: Good point; I'll ditch that unnecessary code and just have this return `''`.

src/utils/web-client.js
tylerconstance: Oh, okay. What needs to change for that to happen?

devinivy: This is something I recall doing together with that `try {} finally {}` block yesterday.

tylerconstance: Oh right! That can be found right here: https://github.com/BigRoomStudios/strangeluv/pull/241/files#diff-273dcc69b8aecc253010766ca5100d5dR44-R51

PR #241, opened 2019-12-02
Allow logging in from a token

tylerconstance:
Login stores a token in local storage. When refreshing or revisiting the view, the app checks if a token exists in user storage and pulls user information from `/api/user` if the token is valid.

### DIFF
--- a/src/middle-end/auth/selectors.js
+++ b/src/middle-end/auth/selectors.js
@@ -12,7 +12,7 @@ exports.getIsAuthenticated = ({ auth }) => {
     // instead of checking for the payload shape
     const isLoginPayload = (payload) => {
 
-        return ['email', 'password'].every((key) => Object.keys(payload).includes(key));
+        return !!payload.token || ['email', 'password'].every((key) => Object.keys(payload).includes(key));
     };
 
     if (!isLoginPayload(index.original || {})) {
@@ -20,4 +20,21 @@ exports.getIsAuthenticated = ({ auth }) => {
     }
 
     return !!(index.result && !index.error);
+};
+
+exports.getToken = ({ auth }) => {
+
+    const { [LOGIN.BASE]: index } = auth.indexes;
+    const { token } = index.result || {};
+
+    if (typeof token !== 'undefined') {
+
+        if (!token) {
+            const error = new Error('No login token on init');
+            error.code = 'NO_TOKEN_ON_INIT';
+            throw error;
+        }
+
+        return token;
+    }
 };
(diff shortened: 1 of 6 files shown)


src/middle-end/auth/selectors.js
```diff
@@ -21,3 +21,20 @@ exports.getIsAuthenticated = ({ auth }) => {
...
 };
+
+exports.getToken = ({ auth }) => {
+
+    const { [LOGIN.BASE]: index } = auth.indexes;
+    const { token } = index.result || {};
+
+    if (typeof token !== 'undefined') {
+
+        if (!token) {
```
devinivy: This looks like code that might be supposed to live in the login action.  For example, when you first visit the site (and you don't have a token in storage), we want the action to fail immediately with an error like the one here: https://github.com/BigRoomStudios/strangeluv/pull/241/files#diff-dfbba6a3d45354bb54b3cefa957b7eb1R31

src/middle-end/auth/selectors.js
```diff
@@ -12,7 +12,7 @@ exports.getIsAuthenticated = ({ auth }) => {
     // instead of checking for the payload shape
     const isLoginPayload = (payload) => {
 
-        return ['email', 'password'].every((key) => Object.keys(payload).includes(key));
+        return !!payload.token || ['email', 'password'].every((key) => Object.keys(payload).includes(key));
```
devinivy: I wonder if one of these would be more consistent in style:
```suggestion
        return !!(payload.token || (payload.username && payload.password));
```
```suggestion
        return  Object.keys(payload).includes('token') || ['email', 'password'].every((key) => Object.keys(payload).includes(key));
```

devinivy: The former above might not be quite right: username, password, or token all may be falsy but present.  So `keys.includes('x')` or `payload.hasOwnProperty('x')` or `typeof payload.x !== 'undefined` might be good directions to go.

src/utils/check-local-storage.js
```diff
@@ -0,0 +1,14 @@
+const LocalStorageAvailable = () => {
```
devinivy:
```suggestion
module.exports = () => {
```

src/utils/check-local-storage.js
```diff
@@ -0,0 +1,14 @@
...
+        localStorage.setItem(test, test);
+        localStorage.removeItem(test);
+        return true;
+    }
+    catch (e) {
+        return false;
+    }
+};
+
+module.exports = LocalStorageAvailable;
```
devinivy:
```suggestion
```

devinivy (approved):

PR #242, opened 2019-12-06
Create authenticated route

tylerconstance:
Authenticated users will be able to view `/exclusive`. Non-authenticated users will get the `NotFound` component, instead.

### DIFF
--- a/src/middle-end/auth/selectors.js
+++ b/src/middle-end/auth/selectors.js
@@ -19,7 +19,8 @@ exports.getIsAuthenticated = ({ auth }) => {
         return false;
     }
 
-    return !!(index.result && !index.error);
+    return !!(index.result && !index.result.loggedOut && !index.error);
+
 };
 
 exports.getHasAuthenticationSettled = ({ auth }) => {
--- a/src/middle-end/auth/selectors.js
+++ b/src/middle-end/auth/selectors.js
@@ -22,6 +22,13 @@ exports.getIsAuthenticated = ({ auth }) => {
     return !!(index.result && !index.error);
 };
 
+exports.getHasAuthenticationSettled = ({ auth }) => {
+
+    const { [LOGIN.BASE]: index } = auth.indexes;
+
+    return Boolean(index && (index.result || index.error));
+};
+
 exports.getToken = ({ auth }) => {
 
     const { [LOGIN.BASE]: index } = auth.indexes;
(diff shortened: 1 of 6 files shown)


src/routes/auth.js
```diff
@@ -0,0 +1,12 @@
+const M = require('../middle-end');
+const { default: connectedAuthWrapper } = require('redux-auth-wrapper/connectedAuthWrapper');
+const NotFoundPage = require('../components/NotFoundPage');
+
+exports.authenticate = connectedAuthWrapper({
+    wrapperDisplayName: 'UserIsAuthenticated',
+    authenticatedSelector: (state) => {
```
devinivy: I expect this will flicker the 404 route while the initial login action completes.  Should be able to address this by configuring `authenticatingSelector`.

tylerconstance: That's totally right. In other projects, we leverage `strange-auth`'s statuses for that selector, but I'll see what we can do here.

src/middle-end/auth/selectors.js
```diff
@@ -22,6 +22,13 @@ exports.getIsAuthenticated = ({ auth }) => {
     return !!(index.result && !index.error);
 };
 
+exports.getHasAuthenticationSettled = ({ auth }) => {
+
+    const { [LOGIN.BASE]: index } = auth.indexes;
+
+    return Boolean(index && (index.result || index.error));
```
devinivy, line 30: After successful logout this will return `false` again, since the logout action doesn't return anything (i.e. `index.result` is null/undefined).

tylerconstance: Oh, that’s that you were trying to tell me in your other repo. What should logout return to address this condition? Just a `result` of `true` or something?

devinivy: I returned `{ loggedOut: true }`, and then had to update the `getIsAuthenticated()` selector to pay attention to that.  The approach gets a little messy handling all these cases, but it works!

jdonisvitch:
Just making note that tests are failing due to an issue with the dispatch and the middle end?

src/middle-end/auth/selectors.js
tylerconstance: [I think this takes care of it](https://github.com/BigRoomStudios/strangeluv/pull/242/commits/507b979867f6d76f99f3591af048d60ddff8f62c), but let me know if I'm still missing something ✨

devinivy: That LGTM!

devinivy:
@jdonisvitch yeah, there's a (newly-)known issue with testing the app when it's connected to the middle-end.  @bmleight has suggested jest module mocking, which seems like the best way forward for the time being!  We're going to handle that independently of these PRs (as an exception, since the bug is with the test suite).

devinivy (approved):

PR #243, opened 2019-12-20
Update pkg 12 2019

jdonisvitch:


devinivy:
Can you remove the package-lock?  We're hoping folks will generate one for themselves when they first install strangeluv.

devinivy (approved):

PR #248, opened 2020-03-17
March maintenance: update deps, fix tests

devinivy:
- [x] Update the tests to deal with middle-end usage issues uncovered in #239.
 - [x] Run of the mill dependency updates.  Worth noting there was one downstream update to strange-middle-end to explicitly support immer v6 (https://github.com/BigRoomStudios/strange-middle-end/commit/a31f1bec560a3e6c6e4dd431dcd43ba953e0b68a).


wswoodruff:
I checked #239 and couldn't easily find the issues uncovered there — but I assume `regenerator-runtime` had something to do with it? Or at least I know I don't understand what `regenerator-runtime` does

wswoodruff (approved):
I don't totally understand what `regenerator-runtime` is for but the rest of these changes look good!

🌵🍞⭕️

src/middle-end/config.js
```diff
@@ -9,12 +9,14 @@ module.exports = {
     }),
     createStore: (reducer, { router }) => {
 
+        const middleware = [
+            MiddleEnd.middleware.thunk,
+            process.env.NODE_ENV !== 'test' && MiddleEnd.middleware.errorLogger,
```
bmleight, line 14: Do we want to include this in a production build?

devinivy: The more I thought about it, the less I saw an issue with it.  If we want to remove it from production, no problem, though!

bmleight: This was just my first thought.  After _actually_ thinking about it and looking at your docs for `errorLogger`, I don't have a problem with it.

src/App.test.js
```diff
@@ -1,12 +1,28 @@
...
 const Testing = require('@testing-library/react');
-const MiddleEnd = require('strange-middle-end');
-const Config = require('./middle-end/config');
-const App = require('./App');
+
+const createSandbox = () => {
+
+    const sandbox = {};
+
+    jest.isolateModules(() => {
```
bmleight: Any test that involves `./middle-end` will need this chunk of code, correct?  If so, should we pull `createSandbox` out into a util of sorts?

devinivy:
Should be good for another peek.  Feel free to merge if it looks good!  We might also want to merge the new changes up to user-recipe once approved just to make sure everything is in sync :)

bmleight (approved):

PR #249, opened 2020-03-17
User recipe maintenance, fix tests

devinivy:
- [x] Bring the user recipe up to date with happenings on the base branch, including readme and dependency updates.
 - [x] Update/fix the tests to deal with middle-end usage issues uncovered in #239.


wswoodruff (approved):
Slight differences here vs #248 — looks good!

🐗
🦀

PR #251, opened 2020-06-16
Rewrites counter container with hooks

tylerconstance:
Couldn't have done it without @zemccartney ✨ 

Replaces the `Counter`'s use of `Connect()` with React Hooks.

Tests are probably failing; this was developed by linking https://github.com/BigRoomStudios/strange-middle-end/pull/12


src/App.js
```diff
@@ -7,30 +7,34 @@ const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiT
 const { default: MuiThemeProvider } = require('@material-ui/styles/ThemeProvider');
 const { default: CssBaseline } = require('@material-ui/core/CssBaseline');
 const StrangeRouter = require('strange-router');
+const MiddleEnd = require('strange-middle-end');
```
tylerconstance, line 10: Do we like `MiddleEnd`, or should we call this something else, like `ReactMiddleEnd`?

devinivy: Looks fine to me!

src/App.js
```diff
@@ -7,30 +7,34 @@ const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiT
...
 const { ConnectedRouter } = require('connected-react-router');
 const { default: ErrorBoundary } = require('react-error-boundary');
 const ErrorFallback = require('./components/ErrorFallback');
 const Routes = require('./routes');
 
-module.exports = ({ store, theme = CreateMuiTheme(), Router = ConnectedRouter, ...routerProps }) => {
+module.exports = ({ middleEnd, store, theme = CreateMuiTheme(), Router = ConnectedRouter, ...routerProps }) => {
```
devinivy: You can omit the `store` here, and just reference `middleEnd.store` below.

tylerconstance: Neat!

src/routes/counter/containers/CounterPage.js
```diff
@@ -1,17 +1,21 @@
...
-    }),
-    {
-        increment: () => M.actions.counter.increment({ amount: 1 }),
-        double: () => M.actions.counter.double()
-    }
-);
+    const m = useMiddleEnd();
+    const counter = useSelector(m.selectors.counter.getValue);
+    const handleIncrement = useCallback(() => m.dispatch.counter.increment({ amount: 1 }), [m]);
+    const handleDouble = useCallback(() => m.dispatch.counter.double(), [m]);
```
devinivy: These `useCallback()`s are are only helping you if `CounterPage` is a pure component or otherwise cares about whether the functions that it is passed change over time.  Here's a great article on it: https://kentcdodds.com/blog/usememo-and-usecallback

devinivy:
This is great!

src/App.js
```diff
@@ -7,30 +7,34 @@ const { default: CreateMuiTheme } = require('@material-ui/core/styles/createMuiT
...
+                    </MiddleEnd.Provider>
                 </ErrorBoundary>
             </MuiThemeProvider>
         </Styled.ThemeProvider>
     );
 };
 
 module.exports.propTypes = {
+    middleEnd: T.object,
     store: T.object.isRequired,
```
tylerconstance: When we remove `store` above, we'll remove it from `propTypes` as well — do we want to make `middleEnd` `required`?

Also, do we need more specificity, like `middleEnd: T.shape({ store: T.object })`?

src/App.test.js
```diff
@@ -11,6 +11,7 @@ it('renders without crashing.', () => {
 
     const { getByText } = Testing.render(
         <App
+            middleEnd={M}
             store={store}
```
tylerconstance: Do we want to remove `store` here?

devinivy: Yeah, if it is no longer used by `App` then it should be removed here and other places it was previously passed.

src/App.js
devinivy: I believe a middle-end is required for `App` to function, so 👍 to making `middleEnd` a required prop.  I don't see any harm in explicitly expressing that `store` is a required key too.

src/routes/counter/containers/CounterPage.js
tylerconstance: Interesting article! Does this mean we could just leave these as functions, like this, to avoid the unnecessary calls to `React.useCallback(…)`?
```suggestion
    const handleIncrement = () => m.dispatch.counter.increment({ amount: 1 });
    const handleDouble = () => m.dispatch.counter.double();
```

devinivy: I think that is just fine :) `useCallback()` is fantastic when you need it, but I think we can keep it simple here.

tylerconstance: Added changes here: https://github.com/BigRoomStudios/strangeluv/pull/251/commits/bf7a9ef7fdaf926facf90e2bcd5ff3550359dc10

devinivy (approved):
Looks good to me!  Final step will be to update the strange-middle-end dependency after it's published.

devinivy (approved):
🎸

PR #252, opened 2020-06-19
Export middle-end as factory, not singleton

zemccartney:
Addresses: https://trello.com/c/8cVAP9cH
Ported from @devinivy 's [prior work on fishbowl](https://github.com/devinivy/fishbowl/commit/f4d46a3e023070b6a17852da553f07096a1fc8e3) 🍻 🐟 ❤️


src/App.test.js
```diff
@@ -2,17 +2,19 @@ require('regenerator-runtime/runtime');
 
 const React = require('react');
 const Testing = require('@testing-library/react');
-const Helpers = require('./test-helpers');
+const App = require('./App');
+const MiddleEnd = require('./middle-end');
```
devinivy: I realize it's a little confusing, but maybe this should be `M` instead for consistency.  I see in `src/index.js` it's called `M`, and strange-middle-end is called `MiddleEnd` in e.g. `src/middle-end/counter/actions.js`.

zemccartney: Ay, yea, that's sloppy, thanks for catching that. I'll tidy that up.

At the risk of bike-shedding, what do you think of revising our naming conventions a bit, as follows, to (I hope) clarify some of this naming confusion?

- `StrangeMiddleEnd` sme's exports
- `MiddleEnd` strangeluv's `middle-end` folder e.g. for requiring 
- `M` the initialized application


Alternatively, I can just fix the issue you flagged, call this good in 5 minutes, and we can punt this nagging, arbitrary discussion down/off the road and not burn more time 😄

devinivy: I will share the considerations I had in mind when building fishbowl, then you can take those and do as you please!
 - I used `middleEnd` or `m` for the initialized middle end instance.  Pascal-cased variables are reserved for classes and `require()`'d dependencies, so `M` didn't feel appropriate to me anymore.  As mentioned before, I toyed with `app`, but it created ambiguity with the mod named "app" for general application stuff, which we seem to often use in our projects too.  It's sort of nice to be able to do e.g. `const { webClient } = m.mods.app` for utilities that are configured at runtime, too!
 - I used `MiddleEnd` for strange-middle-end.  I tried `StrangeMiddleEnd` but it felt overly verbose particularly when creating types and actions.
 - I used `M` for `src/middle-end` since it's always `require()`'d (solely at the app root and in tests).  Since it's a factory for the initialized middle-end instance which I was calling `m`, that also felt reasonably appropriate (sort of like the same relationship between a class and instance, e.g. `const car = new Car()`).

zemccartney: Dig it! Thanks for explaining all that, all sounds good, I'll stick with that here 🙏

src/middle-end/counter/index.js
```diff
@@ -1,5 +1,10 @@
...
+module.exports = (m, options) => {
+
+    return {
+        reducer: require('./reducer'),
+        actions: require('./actions')(m),
+        selectors: require('./selectors')
+    };
 };
+
+
```
devinivy:
```suggestion
```

src/middle-end/counter/index.js
```diff
@@ -1,5 +1,10 @@
-module.exports = {
-    reducer: require('./reducer'),
-    actions: require('./actions'),
-    selectors: require('./selectors')
+module.exports = (m, options) => {
+
+    return {
+        reducer: require('./reducer'),
+        actions: require('./actions')(m),
```
devinivy: Not because it's needed, just to create less variation.
```suggestion
        actions: require('./actions')(m, options),
```

devinivy (approved):

src/middle-end/counter/index.js
zemccartney: ah, werd, I waffled on this, then did the dugg. thanks for noting this, fixing up now!

PR #256, opened 2020-08-07
Update material-ui deps

wswoodruff:
Hmm — so `npx npm-check-updates -m` doesn't catch the material updates first of all, and that probably tells me I don't understand the code I'm running with `npm-check-updates` or `-m` loL.

Updating these libs passed tests locally #itworksonmycomputer


wswoodruff:
Ohp, I introduced package-lock in this PR, brb

PR #258, opened 2020-08-07
Update material-ui packages

wswoodruff:
Tests pass so I'm like why not right

Also here's a copy paste of the description from #256 

---
Hmm — so `npx npm-check-updates -m` doesn't catch the material updates first of all, and that probably tells me I don't understand the code I'm running with `npm-check-updates` or `-m` loL.

Updating these libs passed tests locally #itworksonmycomputer

---
Edit: I think the `-m` looks for only major upgrades, and that's why!

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@material-ui/core": "^4.9.0",
-    "@material-ui/styles": "^4.9.0",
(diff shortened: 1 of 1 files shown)


wswoodruff:
loL the original title made me feel weird

wswoodruff:
Update: Works on Travis too

devinivy (approved):

PR #260, opened 2020-09-03
Recipe aws cdk static site deployment

wswoodruff:
This main PR should be up right?


wswoodruff:
Ugh oh yeah it's a recipe duh

PR #263, opened 2020-11-18
Maintenance Oct 2020 - React 17, other goodies

titchimoto:
**Almost more of an RFC than an actual PR, but we had some pretty big dependencies have major version bumps this last maintenance cycle, so I was just curious what it might look like to have `strangeluv` be updated accordingly to some of these new versions!**

Everything was pretty smooth sailing minus a few `webpack` weird things. As of Webpack 5, `webpack` no longer polyfills `process`, (and also `Buffer` I believe too) and so there were some workarounds there. There's some more info here: https://webpack.js.org/migrate/5/#test-webpack-5-compatibility and details can be found further down that page. The solution was one I saw touted a few places (for example, how `next.js` handled it: https://github.com/vercel/next.js/pull/15499/files) but not sure if this is the best approach for `strangeluv`.

Also a couple of little changes that were mostly made from attempting to get the webpack dev server up and running with the new versions of `webpack` and `webpack-cli`, rather than a place of knowledge and experience. 😅  But running locally with HMR seems fine, and creating builds also seemed fine.

Comments more than welcome please & thank you!


wswoodruff:
Updated `webpack-cli` major version, looks like there aren't breaking changes and it supports webpack 4.x.x || 5.x.x so we should be good there!

bmleight (approved):

PR #264, opened 2021-01-29
Maintenance/jan 2021

wswoodruff:
Most of these packages either had minor or patch upgrades awaiting — only 1 or 2 had major upgrades.

Major version upgrades:
- `immer`'s major upgrade includes the breaking change to `always freeze by default, even in production mode` which I think is fine
- `eslint-config-standard-react` was able to go up to `11.x.x` but that major version results in strange linting output like telling us that `React` is an unused variable when it's clear we need it to be in scope when using JSX — lint passes just fine at `10.x.x` which is what we updated to

Question:
I thought caret syntax `^5.0.1` was the same as `5.x.x` but seems not? Lots of versions in this PR are changing to the `x.x.x` syntax — which gives good output after installing and running `ncu`

Here's the before and after:

Screen Shot 2021-01-29 at 2 09 50 PM

Screen Shot 2021-01-29 at 2 29 01 PM

Summary of what's left:
- We're waiting to upgrade to webpack 5 until `error-overlay-webpack-plugin` is upgraded to support it. There's an open issue [here](https://github.com/smooth-code/error-overlay-webpack-plugin/issues/67)

Paths forward that we're holding back on
- There's a [beta v4](https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0) of `webpack-dev-server` that also updates `webpack-dev-middleware` to support webpack 5 so we'll hold off until we can make that move
- `copy-webpack-plugin@7.x.x` requires webpack 5
- `history` has a major version we can upgrade but I thought we talked about holding off on updating this? There are some things we gotta think about, especially features that are removed as breaking changes. See here https://github.com/ReactTraining/history/releases?after=v5.0.0-beta.1 
And see the breaking changes section here https://github.com/ReactTraining/history/releases


devinivy:
> `eslint-config-standard-react` was able to go up to `11.x.x` but that major version  results in strange linting output like telling us that React is an unused variable when it's clear we need it to be in scope when using JSX

I think this can be explained by this: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html.  So perhaps we could move to v11 taking this into account.

> Question: I thought caret syntax ^5.0.1 was the same as 5.x.x but seems not?

These are identical except `^5.0.1` doesn't match `5.0.0` and `5.x.x` does 👍

wswoodruff:
I thought the caret and x syntax were similar and that's what I read too

The head-scratching part comes when I have it back at caret syntax, rm node_modules, do a fresh install, then run `ncu` to find the large list of updates to do. When the deps are in x.x.x syntax it seems to install the latest pkg versions — confusing, not sure why that's happening

I'm on npm -v 6.14.10

wswoodruff:
I looked into updating `eslint-config-standard-react` to 11.x.x and noticed from the link you sent that they mention using `@babel/preset-react` but we use `babel-preset-react-app` which also makes its way into our webpack config — I haven't been able to get it to compile yet after attempting to set

```
[
    '@babel/preset-react',
    { 'runtime': 'automatic' }
]
```

But was unable to get things to compile — also here are some other weird results from the new eslint settings — it doesn't see that we're using elements imported from `internals`??

Screen Shot 2021-02-03 at 9 45 37 AM

bmleight (approved):
This looks good to me and tested out fine.  Found the issues Bill was working through are related upgrading additional dependencies than what is already checked in here.  Going to continue those efforts from a fresh slate.

PR #265, opened 2021-02-12
upgraded eslint-config-standard-react and introduced @babel/preset-react

bmleight:
Does some stuff:

* upgrades `eslint-config-standard-react` and introduces `@babel/preset-react` to play nice with [React 17's new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
* Removes now unneeded `const React = require('react');` statements just used for JSX (see article linked in ☝️)
* Destructure items being required from react, as per their recommendation in the article:

> Change all default React imports (i.e. import React from "react") to destructured named imports (ex. import { useState } from "react") which is the preferred style going into the future. This codemod will not affect the existing namespace imports (i.e. import * as React from "react") which is also a valid style. The default imports will keep working in React 17, but in the longer term we encourage moving away from them.

There are still the following dependencies to update:
Screen Shot 2021-02-12 at 4 25 30 PM

[error-overlay-webpack-plugin upgrading to webpack 5](https://github.com/smooth-code/error-overlay-webpack-plugin/issues/67) is still holding up upgrading the three webpack libraries.

Upgrading history to [v5](https://github.com/ReactTraining/history/releases/tag/v5.0.0) comes with the following breaking changes:

* Removed support for browsers that do not support the HTML5 history API (no pushState)
* Removed relative pathname support in hash history and memory history
* Removed getUserConfirmation, keyLength, and hashType APIs

It looks like the browser support could be the biggest concern, but according to [caniuse](https://caniuse.com/?search=history) and [mozilla](https://developer.mozilla.org/en-US/docs/Web/API/History#browser_compatibility), the history api is widely supported (I am spoiled on my current project and only have to worry about chrome and am a bit out of date on browser support).  It seems like we could try updating it in the next round.

Screen Shot 2021-02-12 at 4 58 07 PM
Screen Shot 2021-02-12 at 4 55 36 PM

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -1,10 +1,10 @@
-const React = require('react');
+const { lazy: Lazy } = require('react');
 const Layout = require('../components/Layout');
 const NotFoundPage = require('../components/NotFoundPage');
 const NotFoundHelpers = require('./helpers/not-found');
 const HomePage = require('./home/components/HomePage');
 
-const CounterPage = React.lazy(() => import('./counter/containers/CounterPage'));
+const CounterPage = Lazy(() => import('./counter/containers/CounterPage'));
 
 module.exports = [
     {
--- a/config/babel.config.js
+++ b/config/babel.config.js
@@ -5,6 +5,9 @@ module.exports = {
         'react-hot-loader/babel'
     ],
     presets: [
-        'react-app'
+        'react-app',
+        ['@babel/preset-react', {
+            'runtime': 'automatic'
+        }]
     ]
 };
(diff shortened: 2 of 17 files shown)


wswoodruff:
Dude thanks for an awesome writeup on this!

wswoodruff (approved):
Looks great! Thanks for slogging thru the eslint woes

src/routes/index.js
```diff
@@ -1,10 +1,10 @@
-const React = require('react');
+const { lazy: Lazy } = require('react');
 const Layout = require('../components/Layout');
 const NotFoundPage = require('../components/NotFoundPage');
 const NotFoundHelpers = require('./helpers/not-found');
 const HomePage = require('./home/components/HomePage');
 
-const CounterPage = React.lazy(() => import('./counter/containers/CounterPage'));
```
devinivy, line 7: It didn't really bug me as-written, but this is cool too!

devinivy (approved):
Lovely

PR #267, opened 2021-04-21
Bump safe deps

zemccartney:
- Upgrades immer (matching `strange-middle-end` bump forthcoming)
- Replaces deprecated eslint loader with new eslint webpack plugin
- `-vvvv` discussion about webpack upgrading issues below; TLDR I think upgrading is safe; does anyone else agree? or did i goof? 🎩 🍷 🐸 

Updates on the blocked dependency upgrades — webpack-5-related and history — as discussed in the prior 2 months ([Feb/Bri](https://github.com/BigRoomStudios/strangeluv/pull/265) and [Jan/Bill](https://github.com/BigRoomStudios/strangeluv/pull/264))

### history

Still blocked, for a couple of reasons:

- Refer to Bri's writeup on the breaking changes involved
- the upgrade is also blocked by `react-router` and therefore `connected-react-router`, as well, as both are incompatible with history v5, still depending on v4, which seems to have caused a good bit of confusion: 
    - https://github.com/supasate/connected-react-router/issues/487
    - https://github.com/ReactTraining/history/issues/804
    - https://github.com/ReactTraining/history/issues/811

I'm not sure the source of these issues; I'm also not sure why React Training didn't upgrade its packages — e.g. `react-router`, `react-router-dom` — to use the new history version. But the [v5 release of history notes](https://github.com/ReactTraining/history/releases/tag/v5.0.0) this version of history "will serve as the foundation for the upcoming release of React Router version 6." So maybe we're looking at more sweeping upgrades to the react router ecosystem down the line and should hold off on upgrading history till those are complete?

### webpack v5

During his stint, Bri wrote:

> [error-overlay-webpack-plugin upgrading to webpack 5](https://github.com/smooth-code/error-overlay-webpack-plugin/issues/67) is still holding up upgrading the three webpack libraries.

Specifically, we can't perform the following upgrades:

- **webpack**: 4.x.x --> 5.x.x
- **copy-webpack-plugin**: ^6.0.3 --> 8.x.x
- **html-webpack-plugin**: 4.x.x --> 5.x.x

This is still the case, but I did some digging into this to see if anything had changed since Bri's work and found the following, random, questionably helpful things:

- In keeping with [this comment](https://github.com/smooth-code/error-overlay-webpack-plugin/issues/67#issuecomment-817327752), I locally performed the above 3 upgrades and was able to start the app; I threw an error in `App.js`, saw the error overlay as expected. I was able to build and serve (`npm run serve`) the app without any apparent issues (though I could _easily_ be missing something)

- `error-overlay-webpack-plugin` lists webpack as a peer dependency, specifying version 4, but I'm unclear why. As far as I can tell, they do so because they depend on [`react-dev-utils`](https://github.com/facebook/create-react-app/tree/master/packages/react-dev-utils)(`create-react-app`'s webpack utilities), which are in flux due to CRA progressing through their own webpack upgrade: https://github.com/facebook/create-react-app/issues/9994
    - By in flux, I mean, `react-dev-utils` aren't given their own, dedicated section or checklist item on CRA's checklist on the above-linked issue, but issues have cropped up ad-hoc from userland (https://github.com/facebook/create-react-app/issues/9994#issuecomment-811289191 and https://github.com/facebook/create-react-app/pull/10198) as `react-dev-util` dependents have experimented with upgrading to webpack 5 and run into problems. Which might explain why `error-overlay-webpack-plugin` didn't cause any issues on upgrading strangeluv's webpack usage? The parts of `react-dev-utils` that plugin uses might be webpack 5 compatible already. Very very much a guess on my part
    - https://github.com/facebook/create-react-app/pull/10198 this work actually addressed an issue in one of the utils that `error-overlay-webpack-plugin` depends on, which may have made it compatible with webpack 5, since `error-overlay-webpack-plugin`'s dependency is set to bring in this change, a patch, automatically

- If I'm right about any of the above, then `error-overlay-webpack-plugin` might not actually need any upgrading. That is, `react-dev-utils` might be already webpack 5 compatible, which would mean no upgrades necessary (other than fixing the package's `peerDependency` on webpack v4)
    - There is an open PR to upgrade some dependencies on `error-overlay-webpack-plugin`: https://github.com/smooth-code/error-overlay-webpack-plugin/pull/72 But you can see most of the work is to upgrade the example code, not the package itself. And the changes to the actual dependencies (root package.json) look to be incorrect (they remove `react-error-overlay` as a dependency, which definitely breaks the package)

So! All that said, do we feel ok about going ahead with the webpack 5 upgrades, perhaps dealing with some flux in local development due to CRA's ongoing webpack upgrading?


wswoodruff:
Idk I guess it's easier to just wait til the community gets its act together w/ this webpack 5 fiasco

wswoodruff (approved):
Looks grood!

PR #268, opened 2021-05-28
Updates some dependencies

tylerconstance:
The blockers present in @zemccartney's writeup in https://github.com/BigRoomStudios/strangeluv/pull/267 are still present, though it looks like there's an open PR for that issue now on that project, so hopefully we'll be able to update `webpack` in another maintenance cycle or two.

Running `npx npm-check-updates -m` resulted in this list:

```
 babel-jest            26.x.x  →   27.x.x     
 copy-webpack-plugin   ^6.0.3  →   ^9.0.0     
 dotenv                ^8.2.0  →  ^10.0.0     
 html-webpack-plugin    4.x.x  →    5.x.x     
 jest                  26.x.x  →   27.x.x     
 webpack                4.x.x  →    5.x.x     
 history              ^4.10.1  →   ^5.0.0   
 ```
 
`dotenv` managed to get [two major releases](https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md#1000-2021-05-20) in about two weeks, but the only listed breaking change was dropping support for an older version of node that we already don't support in strangeluv.

`jest` was _slightly_ trickier because it had [a significant changelog for version 27](https://github.com/facebook/jest/releases/tag/v27.0.0). Our simple test didn't initially pass, but changing the `testEnvironment` to a newer version which supported `getVmContext` seemed to do the trick.

We were using `jest-environment-jsdom-fifteen`, and searching for the next logical iteration revealed that `jest-environment-jsdom-sixteen` was actually [deprecated](https://github.com/SimenB/jest-environment-jsdom-sixteen#jest-environment-jsdom-sixteen) after being integrated into jest v26. Just changing that string from `jest-environment-jsdom-fifteen` to `jest-environment-jsdom` seemed to do the trick, though, and our test passes again.

Updating `babel-jest` to match, now we're left with the un-upgraded list we're familiar with:
```
 copy-webpack-plugin   ^6.0.3  →  ^9.0.0     
 html-webpack-plugin    4.x.x  →   5.x.x     
 webpack                4.x.x  →   5.x.x     
 history              ^4.10.1  →  ^5.0.0  
 ```


package-lock.json
```diff
@@ -1648,19 +1648,41 @@
       "dev": true
     },
     "@jest/console": {
-      "version": "26.6.2",
-      "resolved": "https://registry.npmjs.org/@jest/console/-/console-26.6.2.tgz",
-      "integrity": "sha512-IY1R2i2aLsLr7Id3S6p2BA82GNWryt4oSvEXLAKc+L2zdi89dSkE8xC1C+0kpATG4JhBJREnQOH7/zmccM2B0g==",
+      "version": "27.0.1",
```
tylerconstance: I wasn't actually expecting to see `package-lock.json` in this project since this is usually the starting point for most projects — should I remove this file in this PR?

devinivy: I think that's a good idea 👍

PR #270, opened 2022-02-10
Maintenance February 2022

tylerconstance:
I noticed an issue after spinning up the latest version of `strangeluv` where I couldn’t interact with any of the buttons or select any text. Some digging revealed this to be an issue caused by the `6.0.10` release of `react-error-overlay` (strangeluv uses `error-overlay-webpack-plugin` and `react-dev-utils`, which both have `react-error-overlay` as a dependency).

Updates:

- Adds `react-error-overlay` `6.0.9` as a dev dependency to avoid a bug caused when updating to `6.0.10`
- Bumps `dotenv` to `^16` ([changelog](https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md#1600-2022-02-02))
- Updates `error-overlay-webpack-plugin` to `v1`

Other notes:

- We’re putting off the `react-router` `v6` update until we update [strange-router](https://github.com/BigRoomStudios/strange-router)
- `eslint-webpack-plugin` drops support for `webpack` `v4` in `3.0.0` ([changelog](https://github.com/webpack-contrib/eslint-webpack-plugin/blob/master/CHANGELOG.md#300-2021-07-19))
- `html-webpack-plugin` also drops support for `webpack` `v4` in `5.0.0` ([changelog](https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md#500-2021-02-03))
- `webpack-dev-server` _will work_ with `webpack` `4.37`, but recommends `v5`, so I figured I'd put that upgrade off for now too ([changelog](https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md#-breaking-changes))
- Putting off `eslint` upgrade to `v8` until `eslint-config-standard-react` supports it (https://github.com/standard/eslint-config-standard-react/issues/68)

There's a lot more information about `webpack` `v5` blockers in the writeup for https://github.com/BigRoomStudios/strangeluv/pull/267, but here's the latest list of major releases we haven't updated to yet:

```
 history                ^4.10.1  →   ^5.2.0     
 react-router-dom         5.x.x  →    6.x.x     
 copy-webpack-plugin     ^6.0.3  →  ^10.2.4     
 eslint                   7.x.x  →    8.x.x     
 eslint-webpack-plugin    2.x.x  →    3.x.x     
 html-webpack-plugin      4.x.x  →    5.x.x        
 webpack                  4.x.x  →    5.x.x     
 webpack-dev-server       3.x.x  →    4.x.x  
 ```


bmleight (approved):
pulled it down for a 🚗  test drive 🚗  and it went👌

PR #271, opened 2022-04-27
Maintenance April 2022

marlonbrs:
Updates:

`error-overlay-webpack-plugin` version pinned to 1.0.0 due to some incompatibility issues v1.1.0 has with webpack-dev-server v3 (https://github.com/gregberge/error-overlay-webpack-plugin/pull/88#discussion_r809890516)

`@mui/material, @mui/styles, @mui/styled-engine-sc` packages added. As part of installing @mui/styled-engine-sc, some changes were made in` webpack.config.js` to replace mui’s default style engine (which is emotion) (https://mui.com/material-ui/guides/styled-engine/)  

`@material-ui/core and @material-ui/styles` have been removed 

Small changes: 
Removed `color` prop from buttons where the value is primary since primary is now the default color value (https://mui.com/material-ui/guides/migration-v4/#button)

`ThemeProvider` is now imported from @mui/material/styles. Though this won’t really have much of an effect to how we style components, I thought going with their suggestion might be useful for others who might use strangeluv (https://mui.com/material-ui/guides/migration-v4/#theme-structure)

Refactored the way `Amber` color was being imported. Material ui mentioned we can’t	import color from `@mui/material/colors/amber` saying nested imports like this are private (only applies for colors though). 
(https://mui.com/material-ui/guides/migration-v4/#mui-material-colors)


noah-green (approved):

noah-green:
merging as part of April maintenance.

PR #273, opened 2022-05-19
Integrate pallies

tylerconstance:
Handles the middle-end work on top of bringing `recipe-user` up-to-date. Replaces `user-pal` with [pallies](https://github.com/frxnz/pallies).

Supports the following routes:
- `/join`
- `/login`
- `/exclusive` (requires authentication)

This can be tested by setting up or using an existing API + database, or cloning `pallies` and using that as the server.

If cloning pallies, enable `CORS` by adding the following to the `server` block in `manifest.js`:
```
          routes: {
            cors: {
                // Needed for refresh token, esp. in local development
                credentials: true
            }
```

Also make sure the `pallies` `.env` is configured with the correct `TOKEN_SECRET`, database information, and `PORT`.


src/containers/Layout.js
```diff
@@ -1,20 +1,20 @@
...
-    (dispatch) => ({
-        logout: async () => {
+module.exports = function LayoutContainer(props) {
 
-            const [err, result] = await M.dispatch.auth.logout();
-            return [err, result];
-        }
-    })
-);
+    const M = useMiddleEnd();
```
devinivy: The original upper-case `M` came from the fact that it was `require()`'d.
```suggestion
    const m = useMiddleEnd();
```

src/containers/Layout.js
```diff
@@ -1,20 +1,20 @@
...
 
-module.exports = internals.connect(Layout);
+    const logout = async () => {
+
+        const [err, result] = await M.dispatch.auth.logout();
+        return [err, result];
+    };
+
+    const isAuthenticated = useSelector(M.selectors.auth.getIsAuthenticated);
```
devinivy:
```suggestion
    const logout = () => m.dispatch.auth.logout();

    const isAuthenticated = useSelector(m.selectors.auth.getIsAuthenticated);
```

src/middle-end/app.js
```diff
@@ -0,0 +1,164 @@
...
+
+internals.createClient = (options) => {
+
+    const client = Axios.create({
+        headers: { common: {} },
+        // Needed for refresh token, esp. in local development
+        withCredentials: true,
+        paramsSerializer: (obj) => {
+            // The default serializer turns array params into ?trades[]=A&trades[]=B
+            // which is not compatible with node's querystring parser which wants ?trades=A&trades=B.
```
devinivy:
```suggestion
            // The default serializer turns array params into ?params[]=A&params[]=B
            // which is not compatible with node's querystring parser which wants ?params=A&params=B.
```

src/middle-end/app.js
```diff
@@ -0,0 +1,164 @@
...
+
+            client.interceptors.response.use(null, (error) => {
+
+                if (error.config?.logoutOnFailure) {
+                    (async () => {
+
+                        await m.dispatch.auth.logout({
+                            reauthorize: error.config.reauthorize
+                        });
+                    })();
```
devinivy:
```suggestion
                    m.dispatch.auth.logout({
                        reauthorize: error.config.reauthorize
                    });
```

src/middle-end/auth/index.js
```diff
@@ -1,6 +1,145 @@
...
+            getCurrentUser: ({ model }) => {
+
+                const { [FETCH_CURRENT_USER.BASE]: index } = model.indexes;
+                return index?.result?.user;
+            },
+            getUserRole: ({ model }) => {
+
+                const { [FETCH_CURRENT_USER.BASE]: index } = model.indexes;
+                return index?.result?.user.role;
+            },
```
devinivy:
```suggestion
            getCurrentUser: ({ model }) => {

                const { [FETCH_CURRENT_USER.BASE]: index } = model.indexes;
                return index?.result?.user;
            },
```

src/middle-end/model/index.js
```diff
@@ -0,0 +1,41 @@
...
+        selectors: {},
+        reducer: bothReducers(entityReducer, createReducer({ mutable: true }, {}, {
+            ...Helpers.logoutHandlers((model) => {
+
+                Object.keys(model.indexes).forEach((index) => {
+
+                    if (index === FETCH_CURRENT_USER.BASE    // This index can stay around, as it maintains auth state
+                    ) {
+                        return;
+                    }
```
devinivy:
```suggestion
                    if (index === FETCH_CURRENT_USER.BASE) {
                        return; // This index can stay around, as it maintains auth state
                    }
```

src/routes/join/containers/SignupPage.js
```diff
@@ -1,19 +1,16 @@
...
+module.exports = function SignupPageContainer(props) {
 
-internals.connect = Connect(
-    (state) => ({}),
-    (dispatch) => ({
-        reqCreateAccount: async ( accountInfo ) => {
+    const M = useMiddleEnd();
 
-            const [err, result] = await M.dispatch.auth.createAccount(accountInfo);
+    const reqCreateAccount = async ( accountInfo ) => {
```
devinivy: Can't suggest this since it's interleaved with deleted lines, but something like this would work equally well:
```js
const reqCreateAccount = (accountInfo) => m.dispatch.auth.register(accountInfo);
```

src/routes/login/containers/LoginPage.js
```diff
@@ -1,21 +1,16 @@
...
-internals.connect = Connect(
-    (state) => ({
-        isAuthenticated: M.selectors.auth.getIsAuthenticated(state)
-    }),
-    (dispatch) => ({
-        reqCreateAccount: async ( loginInfo ) => {
+    const M = useMiddleEnd();
 
-            const [err, result] = await M.dispatch.auth.login(loginInfo);
+    const onPressLogin = async ( accountInfo ) => {
```
devinivy: Can't suggest this since it's interleaved with deleted lines, but something like this would work equally well:
```js
const onPressLogin = (accountInfo) => m.dispatch.auth.login(accountInfo);
```

src/routes/login/containers/LoginPage.js
```diff
@@ -1,21 +1,16 @@
...
-const internals = {};
+module.exports = function LoginPageContainer(props) {
 
-internals.connect = Connect(
-    (state) => ({
-        isAuthenticated: M.selectors.auth.getIsAuthenticated(state)
-    }),
-    (dispatch) => ({
-        reqCreateAccount: async ( loginInfo ) => {
+    const M = useMiddleEnd();
```
devinivy:
```suggestion
    const m = useMiddleEnd();
```

src/routes/join/containers/SignupPage.js
```diff
@@ -1,19 +1,16 @@
...
-const M = require('../../../middle-end');
 
-const internals = {};
+module.exports = function SignupPageContainer(props) {
 
-internals.connect = Connect(
-    (state) => ({}),
-    (dispatch) => ({
-        reqCreateAccount: async ( accountInfo ) => {
+    const M = useMiddleEnd();
```
devinivy:
```suggestion
    const m = useMiddleEnd();
```

devinivy (commented):
Great work!  Pretty much just found some minor tidying that could happen, and in some cases applying familiar conventions.

src/middle-end/auth/index.js
```diff
@@ -1,6 +1,135 @@
-module.exports = {
-    actions: require('./actions'),
-    reducer: require('./reducer'),
-    selectors: require('./selectors'),
-    initialize: require('./init')
+const MiddleEnd = require('strange-middle-end');
+const { ROLES } = require('../../utils/constants');
```
tylerconstance: We're actually not using any roles in this example; should we remove them from `utils/constants`, or leave them as a starting place?

devinivy: I think we should nix them if they don't have any functional use here.

PR #274, opened 2022-05-19
Updates recipe user dependencies

tylerconstance:
Updates `recipe-user` and replaces `strange-forms` with `useState`. Some additional improvements that were missed here were handled in the integrate-pallies branch (https://github.com/BigRoomStudios/strangeluv/pull/273)


config/babel.config.js
```diff
@@ -5,6 +5,9 @@ module.exports = {
         'react-hot-loader/babel'
     ],
     presets: [
-        'react-app'
+        'react-app',
+        ['@babel/preset-react', {
+            'runtime': 'automatic'
```
devinivy:
```suggestion
            runtime: 'automatic'
```

src/routes/join/components/SignupPage.js
```diff
@@ -1,93 +1,51 @@
...
 
-    handleSubmit = async (ev) => {
+    const handleSubmit = async (ev) => {
 
         ev.preventDefault();
-        const accountInfo = this.formatFields();
-        this.setState({ isSubmitting: true });
+        const accountInfo = formatFields();
+        setIsSubmitting(true);
         const [err] = await this.props.reqCreateAccount(accountInfo);
```
devinivy: Might want to test this happy path out— I think this would currently throw an error.
```suggestion
        const [err] = await reqCreateAccount(accountInfo);
```

src/routes/index.js
```diff
@@ -1,14 +1,16 @@
-const React = require('react');
+const { lazy: Lazy } = require('react');
 const Auth = require('./auth');
-const Layout = require('../containers/Layout');
+// const Layout = require('../containers/Layout');
+const Layout = require('../components/Layout');
```
devinivy: In case this was leftover from testing things out:
```suggestion
const Layout = require('../containers/Layout');
```

src/routes/index.js
```diff
@@ -1,14 +1,16 @@
...
 const NotFoundPage = require('../components/NotFoundPage');
 const NotFoundHelpers = require('./helpers/not-found');
 const HomePage = require('./home/components/HomePage');
 const SignupPage = require('./join/containers/SignupPage');
-const LoginPage = require('./login/containers/LoginPage');
+// const LoginPage = require('./login/containers/LoginPage');
+const LoginPage = require('./login/components/LoginPage');
```
devinivy: In case this was leftover from testing things out:
```suggestion
const LoginPage = require('./login/containers/LoginPage');
```

src/routes/index.js
```diff
@@ -1,14 +1,16 @@
-const React = require('react');
+const { lazy: Lazy } = require('react');
```
devinivy, line 1: Was this a linting issue, or something else?

src/routes/auth.js
```diff
@@ -1,10 +1,39 @@
...
+// const { default: connectedAuthWrapper } = require('redux-auth-wrapper/connectedAuthWrapper');
+// const NotFoundPage = require('../components/NotFoundPage');
+
+// exports.authenticate = connectedAuthWrapper({
+//     wrapperDisplayName: 'UserIsAuthenticated',
+//     authenticatedSelector: (state) => M.selectors.auth.getIsAuthenticated(state),
+//     authenticatingSelector: (state) => !M.selectors.auth.getHasAuthenticationSettled(state),
+//     FailureComponent: NotFoundPage
+// });
+
```
devinivy:
```suggestion
```

src/routes/auth.js
```diff
@@ -1,10 +1,39 @@
...
+const { connectedRouterRedirect } = require('redux-auth-wrapper/history4/redirect');
```
devinivy, line 1: I'm a little curious where the `connectedAuthWrapper` was falling down (versus the `history4/redirect`).  Since routing is tied to redux, the `connectedAuthWrapper` might be a little more fitting if it's straightforward to make it work.

devinivy (commented):
Good stuff!  I reviewed the app changes and ignored the config changes for now.  I will take a swing at addressing my review here 👍

src/routes/index.js
tylerconstance: This was just to make this branch render, but I properly addressed it in [the next PR](https://github.com/BigRoomStudios/strangeluv/pull/273/files#diff-268a45f8e3c5fba5953593271f7aaed3d91812853a17dea472417fcc17c84bfdL9-L11) when I actually hooked it up and made it work.

I figured we can waterfall that one into this one  to keep things tidy, but it's no big deal either way.

src/routes/index.js
tylerconstance: Just a style choice; we can leave it as `React.lazy` if we prefer that!

src/routes/join/components/SignupPage.js
tylerconstance: Mhm! That's [also addressed here](https://github.com/BigRoomStudios/strangeluv/pull/273/files#diff-19c2c7a4a6b350795c07620ec34a8b7e2d6777c107b7fab30fc98977f1b37c85L41-L42).

src/routes/auth.js
tylerconstance: I'm up for trying whichever; I just pulled this line-for-line from an implementation in another project. I can point you to which one offline if you're curious! I think you even made the latest commit on that file 😊

marlonbrs (approved):
Did some tests based on the instructions provided [here](https://github.com/BigRoomStudios/strangeluv/pull/273) and everything seems to be working  🙌 . 
I was wondering if we should do some additional tests after [this](https://github.com/BigRoomStudios/strangeluv/pull/273) got merged mainly because of some package version differences. Can take care of that later though 👍

src/routes/auth.js
marlonbrs: I did some research on the differences between `connectedAuthWrapper` and `connectedRouterRedirect` and from my understanding, `connectedAuthWrapper`'s main purpose is for hiding or rendering different components based on authentication status while `connectedRouterRedirect` is mainly used for redirection which I _think_ is what we're trying to go for here with how we're using `Auth.authenticate()` HOC.

There isn't much description as to what each of these methods do and my research was mostly based on how they were used in the APIs getting started [section](https://mjrussell.github.io/redux-auth-wrapper/docs/Getting-Started/ReactRouter4.html)

PR #277, opened 2022-09-09
Sets up basic password reset

tylerconstance:
Sets up basic password reset with an integration with `pallies`. ~~We're currently submitting our request to `/passwordreset2`, but we can wait for some `pallies` cleanup if we'd prefer not to use that endpoint in our boilerplate.~~

_update: this is now `/reset-password` again as of `pallies@0.7.2`_


src/routes/login/components/LoginPage.js
```diff
@@ -29,6 +29,7 @@ module.exports = function SignupPage({ onPressLogin, isAuthenticated }) {
         setIsSubmitting(false);
         if (!err) {
             // Login and redirect
+            window.location.href = '/exclusive';
```
tylerconstance: I think eventually we'll want to use the `useNavigate()` hook in `react-router-dom`, but I don't _think_ we're using a recent enough version of that yet (I haven't looked into it too hard, besides getting a `useNavigate is not a function` error)

src/middle-end/auth/index.js
```diff
@@ -99,17 +99,17 @@ module.exports = (m) => {
...
+                    await client.post('/forgotpassword', { username });
                 }
             }),
             resetPassword: createAction(RESET_PASSWORD, {
-                handler: async ({ token, hash, password }) => {
+                handler: async (args) => {
 
                     const client = getClient();
-                    await client.post('/reset-password', { token, hash, password });
+                    await client.post('/resetpassword2', args);
```
tylerconstance: Keeping this general (`args`) because when this route can _also_ accept `oldPassword` alongside `newPassword` so the user can reset their password while authenticated (without sending a "forgot password" email). We're not currently handling this on the front end, but we _could_.

README.md
```diff
@@ -96,6 +96,18 @@ Runs a hapi server located in `server/` setup to serve the `build/` directory pl
 ### As a hapi plugin
 The production deployment can also be served as a hapi plugin, located in `server/plugin.js`.
 
+## Notes
```
tylerconstance, line 99: Not sure if this is the best place, but we can move and edit this as necessary

src/middle-end/auth/index.js
tylerconstance:
```suggestion
                    await client.post('/reset-password', { oldPassword, userHash, forgotPasswordToken, newPassword });
```

I'm told this is `/reset-password` again as of `pallies@0.7.2`

_Update_: @zemccartney that's a good idea, I updated my suggestion to replace my generic `args` with what's allowed in the `pallies` payload [here](https://github.com/derekwheee/pallies/blob/develop/lib/routes/auth/resetPassword.js#L14).

README.md
```diff
@@ -96,6 +96,18 @@ Runs a hapi server located in `server/` setup to serve the `build/` directory pl
...
+
+### Forgot Password & Password Reset
+#### Integrating with `pallies`
+Internally, we use [pallies](https://github.com/frxnz/pallies) to handle user management and authentication in `hapi`. `pallies` can be used as either a plugin (recommended), or the project can be cloned and be used as its own app server.
+
+When used as a plugin, the application API should implement its own forgot password route, which should call `services().authService.forgotPassword(username)` with the relevant `username`  (e.g., the email address of the user that forgot their password). The service returns a `userHash` and `forgotPasswordToken`, which we'll need to reset the user's password.
+
+The reset password route on the front end has been written to expect `user` and `token` as query params, which means at this stage, we'll likely want to use something like `nodemailer` to send an email with a formatted link to `myapp.com/reset-password?token=myToken&user=myUserHash`. In development, you may wish to user a service like [ethereal](https://ethereal.email/) to capture emails.
```
marlonbrs, line 107:
```suggestion
The reset password route on the front end has been written to expect `user` and `token` as query params, which means at this stage, we'll likely want to use something like `nodemailer` to send an email with a formatted link to `myapp.com/reset-password?token=myToken&user=myUserHash`. In development, you may wish to use a service like [ethereal](https://ethereal.email/) to capture emails.
```

🙏

src/middle-end/auth/index.js
```diff
@@ -99,17 +99,17 @@ module.exports = (m) => {
                 }
             }),
             forgotPassword: createAction(FORGOT_PASSWORD, {
-                handler: async ({ email }) => {
+                handler: async ({ username }) => {
 
                     const client = getClient();
-                    await client.post('/forgot-password', { username: email });
+                    await client.post('/forgotpassword', { username });
```
marlonbrs:
```suggestion
                    await client.post('/forgot-password', { username });
```

src/routes/forgot-password/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,86 @@
...
+
+    const disableSubmit = () => {
+
+        return !email;
+    };
+
+    const handleSubmit = async (ev) => {
+
+        ev.preventDefault();
+        const accountInfo = formatFields();
```
marlonbrs, line 22: Out of curiosity since `email` can be accessed directly,  is there a difference between grabbing `{ username: email }` directly vs  calling `formatFields()` for it?

src/middle-end/auth/index.js
zemccartney: I could also see (and kinda lean toward) slicing this as following pallies and documenting all possible args here, leaving it opaque how they'll be used under the hood, such that we at least know what you could possibly pass here.

Worth testing that passing along undefined params in this way actually works / doesn't lead to any issues on the pallies side, of course

src/routes/login/components/LoginPage.js
zemccartney: Ah, so if memory serves, `useNavigate` is from `react-navigation`, not `react-router-dom` , i think?

Would it work to dispatch an action to the router slice of the middle-end? I'm not actually sure the goal here, so probably muddying the waters; what's the `/exclusive` route?

src/routes/reset-password/containers/ResetPasswordPage.js
```diff
@@ -0,0 +1,14 @@
+const { useMiddleEnd } = require('strange-middle-end');
+const ResetPasswordPage = require('../components/ResetPasswordPage');
+const GetQueryParams = require('../../../utils/get-query-params');
+
+module.exports = function ResetPasswordPageContainer(props) {
+
+    const m = useMiddleEnd();
+
+    const { user: userHash, token: forgotPasswordToken } = GetQueryParams();
```
zemccartney, line 9: Could we lean on the `useParams` hook here? https://v5.reactrouter.com/web/api/Hooks/useparams

src/routes/reset-password/components/ResetPasswordPage.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const handleSubmit = async (ev) => {
+
+        ev.preventDefault();
+        const accountInfo = formatFields();
+        setIsSubmitting(true);
+        const [err] = await onPressResetPassword(accountInfo);
+        setIsSubmitting(false);
+        if (!err) {
+            window.location.href = '/login';
```
zemccartney: Could we use the middle end here, too? Maybe by dispatching a `push` action?

src/routes/reset-password/components/ResetPasswordPage.js
```diff
@@ -0,0 +1,85 @@
...
+                />
+                <Box
+                    my={2}
+                >
+                    <Button
+                        type='submit'
+                        variant='contained'
+                        color='primary'
+                        fullWidth
+                        disabled={disableSubmit() || isSubmitting}
```
zemccartney: No need, but could `isSubmitting` live in `disableSubmit`?

src/routes/reset-password/components/ResetPasswordPage.js
```diff
@@ -0,0 +1,85 @@
...
+                    </Button>
+                </Box>
+            </StyledForm>
+        </PageContainer>
+    );
+};
+
+module.exports.propTypes = {
+    onPressResetPassword: T.func.isRequired,
+    isAuthenticated: T.bool.isRequired
```
zemccartney: Is this used anywhere here? or good to ax?

src/middle-end/auth/index.js
tylerconstance: Thanks! It looks like this was updated in `pallies` on November 4 in [#13](https://github.com/derekwheee/pallies/pull/13/files#diff-0e2753bf7d530e897510c08c4573cd4e163346b43dda9460eb31b39fcdd82260R5)

src/routes/login/components/LoginPage.js
tylerconstance: `/exclusive` is just my questionable name for an authenticated route, only available to users who have signed in. It was a name I felt clever about two years ago for https://github.com/BigRoomStudios/strangeluv/pull/242 and have maintained here for consistency, but I'm very, very open to changing this to better suggestions (`/protected`, `/authenticated`, etc).

`react-router-dom` has a [useNavigate hook](https://reactrouter.com/en/main/hooks/use-navigate) while `react-navigation` has [useNavigation](https://reactnavigation.org/docs/use-navigation/), but I think we're still using `strange-router` for `strangeluv` in lieu of `react-navigation`, so these waters are already pretty swampy, if you catch my drift.

Maybe `push` via `middle-end` is the way to go, like you mentioned in [another suggestion](https://github.com/BigRoomStudios/strangeluv/pull/277/files/a9897e54b0938408c25906bf37e52789c344a8e8#r1050210583), or we could also wait until `react-router-dom` is updated to `6.X`

src/routes/forgot-password/components/ForgotPasswordPage.js
tylerconstance: No difference; just adding an extra abstraction via a little helper function. I agree it looks silly when there's only one, simple field, but it can make things more readable when there's lots of state and mutations that you don't want to interfere with submission logic.

src/routes/reset-password/components/ResetPasswordPage.js
tylerconstance: Sure, why not ✨

src/routes/forgot-password/components/ForgotPasswordPage.js
marlonbrs: That makes sense 🙏 Thank you!

src/routes/reset-password/containers/ResetPasswordPage.js
tylerconstance: Sadly, `useParams` is for _URL_ parameters, which are different than query parameters. There's [an example in react-router that involves query params](https://v5.reactrouter.com/web/example/query-parameters), and it just recommends `URLSearchParams`.

I pulled `query-string` ([npm](https://www.npmjs.com/package/query-string)) in as a dependency for [get-query-params](https://github.com/BigRoomStudios/strangeluv/blob/pallies-forgot-password/src/utils/get-query-params.js) to match another project, but we could use `URLSearchParams` like in that linked example to remove that dependency if you think it's worth it.

README.md
bmleight: I think it's fine to have this here, but wonder if there is a place in nuclino for this type of documentation as well.  And, it might be nice to add your notes from https://github.com/BigRoomStudios/strangeluv/pull/273#issue-1242062595.

README.md
bmleight: My approach was to change the forgotPassword route to log the user hash and token, and then just manually construct the reset password url 🤷

src/routes/forgot-password/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,86 @@
...
+module.exports = function ForgotPasswordPage({ onPressForgotPassword }) {
+
+    const [email, setEmail] = useState('');
+    const [isSubmitting, setIsSubmitting] = useState(false);
+    const [didEmailSend, setDidEmailSend] = useState(false);
+
+    const disableSubmit = () => {
+
+        return !email;
+    };
```
bmleight:
```suggestion
    const isSubmitDisabled = !email || isSubmitting;
```

src/routes/forgot-password/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,86 @@
...
+                />
+                <Box
+                    my={2}
+                >
+                    <Button
+                        type='submit'
+                        variant='contained'
+                        color='primary'
+                        fullWidth
+                        disabled={disableSubmit() || isSubmitting}
```
bmleight: github is not letting me make a suggestion here for some reason, but if you like the suggestion above, this would become: `disabled={isSubmitDisabled}`.  Mainly, I just liked consolidating the logic for if this button is disabled into one place.

src/routes/forgot-password/components/ForgotPasswordPage.js
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    const handleSubmit = async (ev) => {
+
+        ev.preventDefault();
+        const accountInfo = formatFields();
+        setIsSubmitting(true);
+        const [err] = await onPressForgotPassword(accountInfo);
+        setIsSubmitting(false);
+        if (!err) {
```
bmleight, line 26: should we surface this error to the user somehow?  I noticed login and register also don't handle errors, so maybe we could split this off and tackle them all at once.  Could be a good time to bring in snackbar functionality as well.

src/routes/login/components/LoginPage.js
bmleight: [redirect](https://reactrouter.com/en/main/fetch/redirect#redirect) is another option, which they recommend in this case:

> It's recommended to use redirect in loaders and actions rather than useNavigate in your components when the redirect is in response to data.

I'd opt for that or dispatching a push from the middle-end

README.md
tylerconstance: I did the exact same thing you did. I should write that in.

src/routes/forgot-password/components/ForgotPasswordPage.js
tylerconstance: This is a great idea—we already had a card in our development board to handle snackbar integration and some other auth improvements (which I apparently personally made in 2019 😅). The good news is we're doing a good amount of that work here, so I'm repurposing it to cover snackbar integration and error handling :+1:

src/routes/login/components/LoginPage.js
tylerconstance: `redirect` seems like the way to go @bmleight, but we haven't upgraded to `react-router-dom` 6.x yet (it's kind of tied up with the React 18 upgrade we'll get to eventually), so we don't have access to it yet.

I don't remember exactly _why_ that is (it looks like it requires React `v16.8` or greater), and there is also an option to run both [React Routers 5 & 6 together](https://reactrouter.com/en/main/upgrading/v5#backwards-compatibility-package) for an incremental upgrade if we want to badly enough.

bmleight: gah!  I thought I looked, but probably just saw the `Redirect` component and didn't put it all together.  I'd probably opt for [dispatching a push](https://github.com/BigRoomStudios/strangeluv/blob/1b2ffd71a8d9908a23bf9ce990ee627873dd2b8b/src/middle-end/router.js#L15) from the middle end for now.  I also don't remember what was holding us up on the upgrade, but faintly remember an issue with upgrading history.  Probably time to revisit those upgrades.

src/routes/reset-password/components/ResetPasswordPage.js
bmleight: ➕ 1️⃣

src/routes/forgot-password/components/ForgotPasswordPage.js
derekwheee: This is handled in #289 , marking resolved

derekwheee (approved):

PR #281, opened 2022-11-22
Recipe user nov maintenance

tylerconstance:
Updates `recipe-user` with `nov-2022-maintenance` and `update-babel` dependency upgrades.


noah-green (approved):
Runs for me, no errors!

PR #282, opened 2022-11-22
Updates babel

tylerconstance:
@BritneyAHanna and I ran into a little friction upgrading `eslint` and `eslint-config-standard-react` as part of `nov-2022-dev-maintenance`, so I did a little extra digging to move from `babel-eslint`, which had a final update in 2020, to the new monorepo location, `@babel/eslint-parser`.


noah-green (approved):
Runs for me, good stuff!

PR #285, opened 2023-01-19
add jsx quotes rule

marlonbrs:
Trello: https://trello.com/c/c5LAMw92

**Not sure if this would be a problem if you don't have Eslint extension

If you get this error:

Screen Shot 2023-01-19 at 4 20 19 PM

this error won't cause the app to crash or anything but it causes Eslint to not work properly.

What fixed it for me was setting `eslint.nodeEnv` in my vscode extension settings to `development`

### DIFF
--- a/config/eslint.config.js
+++ b/config/eslint.config.js
@@ -31,7 +31,8 @@ module.exports = {
         'react/jsx-uses-vars': 'warn',
         'react/react-in-jsx-scope': 'off',
         'react-hooks/exhaustive-deps': 'warn',
-        'react-hooks/rules-of-hooks': 'error'
+        'react-hooks/rules-of-hooks': 'error',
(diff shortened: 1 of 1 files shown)


bmleight (approved):
looks good!  And thanks for the tip on the eslint plugin.  I had seen that recently when starting up a fresh strangeluv project for kegbot, but hadn't taken time to look into it yet because I was kegbotting.  I do wonder why it wasn't popping up on other strangeluv-based projects though 🤔

marlonbrs:
Not entirely sure if `@babel/eslint-parser`'s got something to do with it? Looking at the other projects, some were still using `babel-eslint` so maybe this new one requires some kind of changes in config but I haven't figured that out yet 🙏

PR #288, opened 2023-02-09
Pallies patch

bmleight:
brings in latest changes from dev maintenance, and updates to work with [pallies v0.7](https://github.com/derekwheee/pallies/pull/13)


src/middle-end/auth/index.js
```diff
@@ -63,19 +63,15 @@ module.exports = (m) => {
...
                     }, {
                         reauthorize: false
                     });
 
-                    if (results.data.accessToken) {
-                        Storage.setItem('accessToken', results.data.accessToken);
-                    }
-
-                    await m.dispatch.auth.fetchCurrentUser();
+                    await m.dispatch.auth.login({ username, password });
```
bmleight, line 74: UX-wise, I'm not sure if we want to automatically log someone in after registering, but found pallies does not return tokens from the register route, so at the very least, we'd remove storing the token.

src/routes/login/components/LoginPage.js
```diff
@@ -82,8 +82,7 @@ module.exports = function LoginPage({ onPressLogin, isAuthenticated }) {
 };
 
 module.exports.propTypes = {
-    onPressLogin: T.func.isRequired,
-    isAuthenticated: T.bool.isRequired
```
bmleight, line 89: this prop was throwing a warning because sometimes it is undefined, and found it's not used, so gave it the 🥾

src/middle-end/auth/index.js
tylerconstance: I like the flow of logging someone in after registering them, just so that after inputting a username and password, the user isn't just redirected to a new page and asked to enter their username and password. I'm open to other perspectives, though!

tylerconstance (approved):

derekwheee (approved):

